PR #24, opened 2023-04-27
Add format example in readme code

laodouya:
- Add format example comment in readme.md code

...

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,14 +43,15 @@ res = chdb.query('select version()', 'Pretty'); print(res.data())
 ### work with Parquet or CSV
 ```python
 # See more data type format in tests/format_output.py
-chdb.query('select * from file("data.parquet", Parquet)', 'JSON')
+res = chdb.query('select * from file("data.parquet", Parquet)', 'JSON'); print(res.get_memview().tobytes().decode("utf-8"))
+# With large data result, get_memview() is more quickly cos it does not do an extra data copy.
 res = chdb.query('select * from file("data.csv", CSV)', 'CSV');  print(str(res.get_memview().tobytes()))
 ```
 
 ### pandas dataframe output
 ```python
-# input type and output type support may different
-# see more in https://clickhouse.com/docs/en/interfaces/formats
+# Input type and output type support may different.
+# See more in https://clickhouse.com/docs/en/interfaces/formats
 chdb.query('select * from file("data.parquet", Parquet)', 'Dataframe')
 ```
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -37,17 +37,20 @@ Currently, chDB only supports `query` function, which is used to execute SQL and
 
 ```python
 import chdb
-res = chdb.query('select version()', 'CSV'); print(str(res.get_memview().tobytes()))
+res = chdb.query('select version()', 'JSON'); print(str(res.get_memview().tobytes()))
```
lmangani, line 40: Not sure if JSON is a good default example from a performance and response size / readability point of view.

README.md
```diff
@@ -37,17 +37,20 @@ Currently, chDB only supports `query` function, which is used to execute SQL and
...
-res = chdb.query('select version()', 'CSV'); print(str(res.get_memview().tobytes()))
+res = chdb.query('select version()', 'Pretty'); print(res.data())
 ```
 
 ### work with Parquet or CSV
 ```python
-chdb.query('select * from file("data.parquet", Parquet)', 'CSV')
-chdb.query('select * from file("data.csv", CSV)', 'CSV')
+# See more data type format in tests/format_output.py
+chdb.query('select * from file("data.parquet", Parquet)', 'JSON')
```
auxten, line 46: Also print the query result?

README.md
```diff
@@ -37,17 +37,20 @@ Currently, chDB only supports `query` function, which is used to execute SQL and
...
+res = chdb.query('select * from file("data.csv", CSV)', 'CSV');  print(str(res.get_memview().tobytes()))
 ```
 
 ### pandas dataframe output
 ```python
+# input type and output type support may different
```
auxten, line 52: Cap the first word

README.md
auxten: I think JSON is okay, as it's so widely used

README.md
```diff
@@ -37,17 +37,20 @@ Currently, chDB only supports `query` function, which is used to execute SQL and
 
 ```python
 import chdb
-res = chdb.query('select version()', 'CSV'); print(str(res.get_memview().tobytes()))
+res = chdb.query('select version()', 'Pretty'); print(res.data())
```
auxten, line 40: Not intent to let user use `data()` function. As it will do an extra data copy of the result. 
But as it is much shorter than `get_memview().tobytes()`, I think this is ok.
Most Python users care more about less code than little performance difference.

README.md
lmangani: Then perhaps JSONCompact is a good compromise. This is what play UI uses as well.

PR #25, opened 2023-04-27
Add function: chdb() to return version

lmangani:
This PR adds a simple `chdb()` function with no parameters that returns the current version of chDB as a string. 
The version number is defined using the `CHDB_VERSION` macro, which should be defined during the build process.

The function is conditionally compiled using an `#if` directive that checks whether the `CHDB_VERSION_STRING` macro derived from `CHDB_VERSION` is defined, to avoid building and registering the function if the version number is not available.

- New Feature

add function `chdb()` returning the current build version


lmangani:
Tested w/ ClickHouse


Build Flag:
```
-DCHDB_VERSION="0.6.2"
```

Validation: 
```
# docker run -ti --rm clickhouse:chdb "SELECT version()"
23.4.1.1
# docker run -ti --rm clickhouse:chdb"SELECT chdb()"
0.6.2
```

PR #44, opened 2023-06-26
Fix nullptr caused core dump

auxten:
```bash
python3 -m chdb "INSERT INTO FUNCTION url('https://urleng.com/xxx', JSONEachRow, 'key String, value UInt64') VALUES ('hello', 2)" Pretty
```
got `Aborted (core dumped)`


auxten:
@all-contributors please add auxten for code

PR #58, opened 2023-07-12
ChDB_TPCH_SF10_Queries

reema93jain:
Hi,

I have moved this folder under tests directory.

Thank you
Reema Jain


lmangani:
Thanks @reema93jain this is great! While @auxten confirms, perhaps you could open an issue about this failure?
> ChDB fails to execute Query 21 of TPCH_SF10

reema93jain:
Thank you @lmangani! Yes, will do so.

PR #70, opened 2023-07-26
Add arm64 linux build

nmreadelf:
Add arm64 build


lmangani:
Wouldn't the regular chdb builder script just work with the arm64 runner?

.github/workflows/build_arm_whells.yml
```diff
@@ -0,0 +1,161 @@
...
...cmYgLi9idWlsZGxpYgoKICAgIC4vY2hkYi9idWlsZC5zaAogICAgY2QgY2hkYiAmJiBweXRob24zIC1jICJpbXBvcnQgX2NoZGI7IHJlcyA9IF9jaGRiLnF1ZXJ5KCdzZWxlY3QgMTExMjIyMjIyMiw1NTUnLCAnSlNPTicpOyBwcmludChyZXMpIiAmJiBjZCAtCgogICAgLi9nZW5fbWFuaWZlc3Quc2gKICAgIGNhdCAuL01BTklGRVNULmluCgogICAgcHl0aG9uMyAtbSBidWlsZCAtLXdoZWVsCgogICAgIyB0cnkgZGVsZXRlIAogICAgcm0gLWYgZGlzdC9jaGRiLSotY3Ake1BZX1ZFUi8vLi99LWNwJHtQWV9WRVIvLy4vfS1saW51eF9hYXJjaDY0LndobCB8fCA6CgogICAgcHl0aG9uMyAtbSB3aGVlbCB0YWdzIC0tcGxhdGZvcm0tdGFnPW1hbnlsaW51eF8yXzI0X2FhcmNoNjQgLS1yZW1vdmUgZGlzdC9jaGRiLSotY3Ake1BZX1ZFUi8vLi99LWNwJHtQWV9WRVIvLy4vfS1saW51eF9hYXJjaDY0LndobAoKICAgIGZpbmQgZGlzdAoKICAgIHB5dGhvbjMgLW0gcGlwIGluc3RhbGwgcGFuZGFzIHB5YXJyb3cgcHN1dGlsCiAgICBmaW5kIGRpc3QKICAgIHB5dGhvbjMgLW0gcGlwIGluc3RhbGwgLS1mb3JjZS1yZWluc3RhbGwgZGlzdC9jaGRiLSotY3Ake1BZX1ZFUi8vLi99LWNwJHtQWV9WRVIvLy4vfS1saW51eF9hYXJjaDY0LndobAoKICAgIHB5dGhvbjMgLWMgImltcG9ydCBjaGRiOyByZXMgPSBjaGRiLnF1ZXJ5KCdzZWxlY3QgdmVyc2lvbigpJywgJ0NTVicpOyBwcmludChyZXMpIgoKICAgIHB5dGhvbjMgLW0gY2hkYiAiU0VMRUNUIDEsICdhYiciIGFycm93dGFibGUKCiAgICBweXRob24zIC1tIGNoZGIgIlNFTEVDVCAxLCAnYWInIiBkYXRhZnJhbWUKCiAgICBiYXNoIC14IC4vY2hkYi90ZXN0X3Ntb2tlLnNoCgogICAgbWFrZSB0ZXN0CmRvbmUK | base64 -d | bash -l -e -x
```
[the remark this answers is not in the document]
lmangani: the current build failure seems caused inside this obfuscated part:
```
rm -f dist/chdb-*-cp${PY_VER//./}-cp${PY_VER//./}-linux_aarch64.whl || :
```

.github/workflows/build_arm_whells.yml
```diff
@@ -0,0 +1,140 @@
...
+  build_wheels_linux_arm64:
+    name: ${{ matrix.os }} py${{ matrix.python-version }}
+    runs-on:
+      - self-hosted
+      - ARM64
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ ubuntu-20.04 ]
+        # python-version: [ "3.8", "3.9", "3.10", "3.11"]
```
lmangani, line 32: if the script runs successfully, shall we enable more versions to be built?

nmreadelf: the script build 3.8 3.9 3.10. 3.11, look at packages/build_linux_arm64.sh line 7th.

lmangani: Should we not use the matrix values for parallel builds?

nmreadelf: Currently our self-hosted runner only supports one parallel job. This may be related to the fact that we only have one self-hosted runner. Also, there are some repetitive operations across different versions, such as downloading ccache and submodule caches. These operations not only consume time but also take up a considerable amount of disk space (currently one job takes up around 5GB, not to mention the intermediate files generated during the build process).

lmangani: Understood. Then perhaps we could at least use an ENV variable as list of the python builds to run, in order for this not to be 100% hardcoded?

lmangani:
Last request I promise! The build script is currently located in`packages/build_linux_arm64.sh` but the ideal location for this file should be to remain alongside the other chdb scripts in the `chdb` folder. What do you think @auxten ?

auxten:
> Last request I promise! The build script is currently located in`packages/build_linux_arm64.sh` but the ideal location for this file should be to remain alongside the other chdb scripts in the `chdb` folder. What do you think @auxten ?

I think we can move it to the chdb directory. I previously placed `[build_mac_arm64.sh](https://github.com/chdb-io/chdb/blob/main/packages/build_mac_arm64.sh)` in the packages folder, but it doesn't seem like a good idea. We can move them all to the chdb directory. Don't forget to update the script paths in the Makefile as well. @nmreadelf

lmangani (approved):

chdb/build_linux_arm64.sh
```diff
@@ -0,0 +1,58 @@
...
+    # if python3 -VV does not contain ${PY_VER}, then exit
+    if ! python3 -VV 2>&1 | grep -q "${PY_VER}"; then
+        echo "Error: Required version of Python (${PY_VER}) not found. Aborting."
+        exit 1
+    fi
+
+    python3 -m pip install -U pybind11 wheel build tox
+    rm -rf ./buildlib
+
+    ./chdb/build.sh
```
lmangani, line 28: I'm watching the runner build. Since the files inside the chdb build folder don't change, maybe this step could be moved ahead of the loop and completed once for all executions to speed up things?

auxten: Python header files changed, so the `_chdb.python.xxx.so` will be different. It's necessary to run `./chdb/build.sh` for every python version.

lmangani: ACK. Will we need a better cleanup strategy to avoid filling up the builder storage.

PR #77, opened 2023-07-30
Broader aarch64 compatibility

lmangani:
Additional build flag for broader compatibility with ARM7/8 aarch64 CPUs for both libchdb bindings and python releases

NOTE: Once merged, [libchdb builder](https://github.com/metrico/libchdb/blob/main/.github/workflows/build_lib_arm64.yml#L85) will need to be updated to assume compatibility


auxten:
- Move `-DNO_ARMV81_OR_HIGHER=1` to CPU_FEATURES
- macOS arm64 support ARMv8.4-A

PR #99, opened 2023-08-26
Strip library/wheels

lmangani:
As we grow in arch coverage and versions, space matters. 

Stripping the library could save 100+ megabytes per release/package and the impact on our ability to troubleshoot should be limited since we can easily rebuild locally as needed or bind this to an ENV variable if required.

- Build/Testing/Packaging Improvement

Strip distribution library used by wheels


auxten (approved):
Sure, it seems the debug info did not help too much.
Let's strip it!

PR #130, opened 2023-11-10
Data input hooks (WIP)

lmangani:
Work in progress for tracking and comments only

### Changelog category:
- New Feature

### Changelog entry
Optional handlers for `--input_format` and data input `--file`


lmangani:
@auxten @laodouya perhaps `--file` could be attached to a FIFO named pipe as well? This way whatever the app pushes to the FIFO ahead of execution, could be processed by the query

PR #159, opened 2023-12-20
Fix misused clang flag on macOS

Daniel-Robbins:
Sorry, I made a mistake. Maybe you need to re-release the 1.1.0 @auxten


auxten:
As the pypi v1.1.0 is not uploaded, I will re-tag and release.
Maybe we can merge after the merge check pipeline succeed next time.
Re-tag is not good. @Daniel-Robbins

PR #161, opened 2023-12-20
Feat: golang sql interface

nmreadelf:


nmreadelf:
close, move to chdb-go

Issue #162, opened 2023-12-20
DBAPI doesn't support CREATE TABLE or other none-returning queries


yunyu:
**Describe what's wrong**

CHDB's dbapi implementation does not work for basic queries like `CREATE TABLE` that return no data. I am using DBAPI for sqlalchemy compatibility. 

**Does it reproduce on recent release?**

Yes, reproduced on `chdb==1.0.2`

**How to reproduce**

```
import chdb.dbapi as dbapi
conn1 = dbapi.connect()
cur1 = conn1.cursor()
cur1.execute("""CREATE TABLE rate (
	day Date, 
	value Int32
) ENGINE = Memory""")
```

**Expected behavior**

Execute succeeds without an exception

**Error message and/or stacktrace**

```
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
File ~/venv/lib/python3.11/site-packages/chdb/dbapi/connections.py:185, in CHDBResult.read(self)
    184 try:
--> 185     data = json.loads(self.connection.resp)
    186 except Exception as error:

File /usr/local/lib/python3.11/json/__init__.py:339, in loads(s, cls, object_hook, parse_float, parse_int, parse_constant, object_pairs_hook, **kw)
    338 if not isinstance(s, (bytes, bytearray)):
--> 339     raise TypeError(f'the JSON object must be str, bytes or bytearray, '
    340                     f'not {s.__class__.__name__}')
    341 s = s.decode(detect_encoding(s), 'surrogatepass')

TypeError: the JSON object must be str, bytes or bytearray, not NoneType

During handling of the above exception, another exception occurred:
...
File ~/venv/lib/python3.11/site-packages/chdb/dbapi/connections.py:187, in CHDBResult.read(self)
    185     data = json.loads(self.connection.resp)
    186 except Exception as error:
--> 187     raise err.InterfaceError("Unsupported response format:" % error)
    189 try:
    190     self.field_count = len(data["meta"])

TypeError: not all arguments converted during string formatting
```

**Additional context**

This seems to work fine in the non DBAPI implementation.

yunyu:
It looks like this behavior is caused by:

https://github.com/chdb-io/chdb/blob/b76b240cc45c96d796c2a4783aa8856a6099c9d0/chdb/dbapi/connections.py#L123

[...]

PR #169, opened 2023-12-21
Make DBAPI implementation use session, fix execute for none-returning queries

### LINKED ISSUE #163: DBAPI implementation loses all Clickhouse state on every query
**Describe what's wrong**

It is impossible to use dbapi for anything involving writes, because it seems like the Clickhouse state is reset on every single query.

**How to reproduce**

```
from chdb import dbapi

conn1 = dbapi.connect()
cur1 = conn1.cursor()

cur1.execute("create database test_db; select 1")
cur1.execute("use test_db")
```
(The select 1 in `CREATE DATABASE` is there due to #162)

**Expected behavior**

`test_db` is created and can be used in the subsequent query

**Error message and/or stacktrace**

`Code: 81. DB::Exception: Database test_db does not exist. (UNKNOWN_DATABASE)`

**Additional information**

This can be easily verified by running `select * from system.databases` in succession:

```
from chdb import dbapi

conn1 = dbapi.connect()
cur1 = conn1.cursor()

cur1.execute("create database test_db; select * from system.databases;")
print(cur1.fetchall())

cur1.execute("select * from system.databases;")
print(cur1.fetchall())
```

You will notice that `test_db` disappears in the second call, and the directory names reflect a different unix timestamp which is probably not correct.

yunyu:
Fixes #163 #162 

Make dbapi implementation use session, which is stateful (most consumers of dbapi likely want statefulness). Also fix `execute` to not throw an exception for None-returning queries like `CREATE TABLE`.

### DIFF
--- a/chdb/dbapi/connections.py
+++ b/chdb/dbapi/connections.py
@@ -23,6 +23,7 @@ class Connection(object):
     """
 
     _closed = False
+    _session = None
 
     def __init__(self, cursorclass=Cursor):
 
@@ -39,6 +40,8 @@ class Connection(object):
         self.connect()
 
     def connect(self):
+        from chdb import session as chs
+        self._session = chs.Session()
         self._closed = False
         self._execute_command("select 1;")
         self._read_query_result()
@@ -55,6 +58,7 @@ class Connection(object):
         if self._closed:
             raise err.Error("Already closed")
         self._closed = True
+        self._session = None
 
     @property
     def open(self):
@@ -119,8 +123,7 @@ class Connection(object):
         if DEBUG:
             print("DEBUG: query:", sql)
         try:
-            import chdb
-            self._resp = chdb.query(sql, output_format="JSON").data()
+            self._resp = self._session.query(sql, output_format="JSON").data()
         except Exception as error:
             raise err.InterfaceError("query err: %s" % error)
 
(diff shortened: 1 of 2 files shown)


lmangani (approved):

yunyu:
Hmm, I signed the CLA but the comment isn't updating:
image

Also not sure why the bot is saying "Yunyu Lin seems not to be a GitHub user"

lmangani:
@yunyu this usually means the commits within the PR were pushed to your branch with a different user (or no user)

yunyu:
Got it, CLA fixed!

yunyu:
Btw @lmangani do you know why the test fails with `ENGINE = Memory` (which should persist data, given it's the same session?) but not with `ENGINE = Log`?

lmangani:
> Btw @lmangani do you know why the test fails with `ENGINE = Memory` (which should persist data, given it's the same session?) but not with `ENGINE = Log`?

Sessions are files. My guess the Memory engine bypasses the session persistence entirely. @auxten can confirm best.

auxten:
@lmangani is right. chDB session relies on persistence of database storage files.
But, supporting session on `ENGINE = Memory` is reasonable. I will find a way to support that.

laodouya (approved):
LGTM

PR #176, opened 2023-12-23
Allow path in dbapi connect

nevinpuri:
Just as you can pass a `path` argument to the `dbapi.connect()` function (enabling persistency). This will allow similar functionality as the `Session` api already provides, but using the dbapi.

- Improvement

Added `path` parameter to dbapi

- Not sure this requires a documentation entry, but here it is:
`path`: the folder path to use for database data on disk.
- [x] Documentation is written (mandatory for new features)


Nevin1901:
Don't merge yet, I just realized this change isn't fully working

Nevin1901:
I've updated all the code. Should work but I haven't tested.

auxten:
Please add tests for your modification.

nevinpuri:
@auxten I have added a test. It all should work now 👍

tests/test_dbapi_persistence.py
```diff
@@ -0,0 +1,27 @@
...
+
+import unittest
+from chdb import dbapi
+
+test_state_dir = ".state_tmp_auxten_dbapi"
+
+class TestDBAPIPersistence(unittest.TestCase):
+    def test_persistence(self):
+        conn = dbapi.connect(path=test_state_dir)
+        cur = con.cursor()
```
auxten, line 11: conn?

auxten:
Fixing some test issue
organizations-2000000.zip link expired. Just use 
[organizations-500000.zip](https://github.com/chdb-io/chdb/files/14662379/organizations-500000.zip)

PR #181, opened 2024-01-08
Disable output_format_json_quote_64bit_integers

### LINKED ISSUE #117: dbapi return int as string
chdb driver version: 0.14.2
chdb version:  23.6.1.1
Python version: 3.9.6
macos catalina: 10.15.7

How to reproduce:
create test table.
```sql
CREATE TABLE t1
(
    `id` Int64,
    c1 String 
)
ENGINE = ReplacingMergeTree
ORDER BY id;

insert into t1 values
(1,'a'),
(2,'b'),
(3,'c');
```

python code, just read data from table t1.
```python
from chdb import dbapi

conn1 = dbapi.connect()
cur1 = conn1.cursor()

cur1.execute("select id, c1 from remote('host:9000','db','table','user','password') final")
result = cur1.fetchall()
print(result)

cur1.close()
conn1.close()
```

when i run the python code as above, i got the below result, the int id return as string, and the result returned as tuple, is it possible dbapi return result same as clickhouse_driver in python?
```
(('1', 'a'), ('2', 'b'), ('3', 'c'))
```

yunyu:
Fixes #117

Makes dbapi (main JSON output format consumer in chdb) return int objects instead of str for int64 columns, since python supports int64 parsing (vs JS for which the default was set)

### DIFF
--- a/src/Core/Settings.h
+++ b/src/Core/Settings.h
@@ -979,7 +979,7 @@ class IColumn;
     M(UInt64, format_binary_max_array_size, 1_GiB, "The maximum allowed size for Array in RowBinary format. It prevents allocating large amount of memory in case of corrupted data. 0 means there is no limit", 0) \
     M(URI, format_avro_schema_registry_url, "", "For AvroConfluent format: Confluent Schema Registry URL.", 0) \
     \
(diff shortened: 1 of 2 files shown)


yunyu:
@auxten can I get a review?

auxten:
LGTM

PR #227, opened 2024-06-18
226 crash on mergetree background task cleanup

### LINKED ISSUE #226: Crash on MergeTree background task cleanup
Crash in chdb-node https://github.com/chdb-io/chdb-node/issues/11
```
terminate called after throwing an instance of 'DB::Exception'
  what():  Storage is about to be deleted. Done active task as if it was aborted.
Aborted (core dumped)
```

Reproduce:

```js
const chdb = require('chdb');

Array.from(Array(10)).forEach(() => {
  const session = new chdb.Session();

  session.query('CREATE DATABASE local');

  session.query(`
  USE local; CREATE TABLE IF NOT EXISTS knowledge_base_portal_interface_event
  (
      timestamp                   DateTime64,
      company_id                  Int64,
      event_type                  String,
      locale                      String,
      article_id                  Int64 DEFAUlT 0,
  )
  ENGINE = MergeTree
  ORDER BY (company_id, locale, timestamp)
  COMMENT 'This table represents a store of knowledge base portal interface events';
  `);

  Array.from(Array(15)).forEach((_x, index) => {
    session.query(
      `USE local; INSERT INTO knowledge_base_portal_interface_event FORMAT JSONEachRow [{"company_id":${index},"locale":"en","timestamp":1717780952772,"event_type":"article_update","article_id":7},{"company_id":${
        index + 100
      },"locale":"en","timestamp":1717780952772,"event_type":"article_update","article_id":7}]`
    );
  });

  ret = session.query('SELECT * FROM knowledge_base_portal_interface_event', 'JSONObjectEachRow');
  console.log("Session Query Result:", ret);
  session.cleanup();
});
```

auxten:
Fix #226 
and
Various fix


yens1018:
@auxten Thank you for the fix. Is it possible to release the package change to npm? We would like to use this to unblock our CI. (Details: https://github.com/chdb-io/chdb-node/issues/11)

auxten:
> @auxten Thank you for the fix. Is it possible to release the package change to npm? We would like to use this to unblock our CI. (Details: [chdb-io/chdb-node#11](https://github.com/chdb-io/chdb-node/issues/11))

The procedure:
1. Test on all platform [Linux,macOS]x[x86_64,arm64]
2. Release libchdb.so and chdb python wheels
3. Update chdb-node to latest libchdb.so and test
4. Release chdb-node

PR #257, opened 2024-08-18
fix CmakeLists comment

DemoYeti:
Simple change to CmakeLists, making the grammar of the comment correct


tylerhannan:
@DemoYeti - thanks for this change.

@auxten - approved!

PR #263, opened 2024-09-03
Try to fix gpt example

xinhuitian:
In examples/chDB_GPT.ipynb, the refine_query part should print the _query_ str instead of _answer_ I think? 

Meanwhile, for many LLMs (I used Zhipu for test in this case), the generated result always contains a markdown code block syntax around the real sql, so maybe we can add two replace operations to remove this? The test result is shown as below:
image


auxten:
Thanks for your contribution

Issue #262, opened 2024-09-02
Which clickhouse engines dose chDB currently support and will support in the future


DanielMao1:
Hello, I would like to know if there is a list to let me know that, amoung so many engines clickhouse provides, which are supported by chDB? As far as I know, [`Memory` engine has some issues](https://github.com/chdb-io/chdb/issues/197), so I would like to know, which are well supported? (**To be more specific, MergeTree, AggregatingMergeTree, Memory are the three types we are now considering to use, I would like to know if they are supported**) 
## Database engines:
[Atomic](https://clickhouse.com/docs/en/engines/database-engines/atomic)
[Lazy](https://clickhouse.com/docs/en/engines/database-engines/lazy)
[Replicated](https://clickhouse.com/docs/en/engines/database-engines/replicated)
[PostgreSQL](https://clickhouse.com/docs/en/engines/database-engines/postgresql)
[MySQL](https://clickhouse.com/docs/en/engines/database-engines/mysql)
[SQLite](https://clickhouse.com/docs/en/engines/database-engines/sqlite)
[MaterializedPostgreSQL](https://clickhouse.com/docs/en/engines/database-engines/materialized-postgresql)
[MaterializedMySQL](https://clickhouse.com/docs/en/engines/database-engines/materialized-mysql)
## Mergetree engines:
[MergeTree](https://clickhouse.com/docs/en/engines/table-engines/mergetree-family/mergetree)
[Data Replication](https://clickhouse.com/docs/en/engines/table-engines/mergetree-family/replication)
[Custom Partitioning Key](https://clickhouse.com/docs/en/engines/table-engines/mergetree-family/custom-partitioning-key)
[ReplacingMergeTree](https://clickhouse.com/docs/en/engines/table-engines/mergetree-family/replacingmergetree)
[SummingMergeTree](https://clickhouse.com/docs/en/engines/table-engines/mergetree-family/summingmergetree)
[AggregatingMergeTree](https://clickhouse.com/docs/en/engines/table-engines/mergetree-family/aggregatingmergetree)
[CollapsingMergeTree](https://clickhouse.com/docs/en/engines/table-engines/mergetree-family/collapsingmergetree)
[VersionedCollapsingMergeTree](https://clickhouse.com/docs/en/engines/table-engines/mergetree-family/versionedcollapsingmergetree)
[GraphiteMergeTree](https://clickhouse.com/docs/en/engines/table-engines/mergetree-family/graphitemergetree)
[...]

Issue #258, opened 2024-08-21
Create table view/ import  from pandas in a session


DanielMao1:
Hello! : ) I am working on some feature engineering tasks and I find your great job. But I encountered some API limitations. Such APIs are supported by clickhouse and duckdb, but not supported by chDB. So I would like to discuss this issue.

**Use case**

When I using clickhouse, my use case is like:

- firstly, I got some pandas dataframes:

```python
import pandas as pd
import clickhouse_connect

# 1. load some dataframes
ads = pd.DataFrame({
    'ad_id': [1, 2, 3, 4, 5],
    'userid': [101, 102, 103, 104, 105]
})

users = pd.DataFrame({
    'userid': [101, 102, 103, 106, 107]
})

pageviews = pd.DataFrame({
    'userid': [101, 102, 103, 101, 105],
    'platform': [1, 2, 1, 2, 1]
})

```

- Then I load them into clickhouse:

```python
# 2. import dataframes into clickhouse
client = clickhouse_connect.get_client(host='localhost')

client.command('DROP TABLE IF EXISTS ads')
client.command('DROP TABLE IF EXISTS users')
client.command('DROP TABLE IF EXISTS pageviews')

client.command('CREATE TABLE ads (ad_id Int64, userid Int64) ENGINE = Memory')
client.command('CREATE TABLE users (userid Int64) ENGINE = Memory')
client.command('CREATE TABLE pageviews (userid Int64, platform Int64) ENGINE = Memory')

client.insert_df('ads', ads)
client.insert_df('users', users)
client.insert_df('pageviews', pageviews)

```

- Then I run some very complex SQLs (which consists of create temp table statements and join 5-6 tables) on the database and get result as dataframes:

```python
# generate temporary results
temp_query = """
SELECT
    ads.ad_id,
    ads.userid
FROM ads
JOIN users ON ads.userid = users.userid
"""

client.command('DROP TABLE IF EXISTS temp_ads_users')
client.command('CREATE TABLE temp_ads_users ENGINE = Memory AS ' + temp_query)

# generate final results
final_query = """
SELECT
    temp_ads_users.ad_id,
    mode() AS most_common_platform
FROM temp_ads_users
JOIN pageviews ON temp_ads_users.userid = pageviews.userid
GROUP BY temp_ads_users.ad_id
"""

result_df = client.query_df(final_query)
print(result_df)
```

**Describe the solution you'd like**

[...]

PR #283, opened 2024-11-02
SQLite like API mode of chDB 

### LINKED ISSUE #197: Reimplement the session mode
Current chDB(till v2.0.2) relys on temp disk storage to keep the session data.
Everytime `session.query` runs almost everything in memory will be recreated and reinit which caused a lot of state problems like:

- [Memory engine could not keep state in session.](https://github.com/orgs/chdb-io/discussions/196)
- https://github.com/chdb-io/chdb/issues/225
- https://github.com/chdb-io/chdb/issues/258
- https://github.com/chdb-io/chdb/issues/261
- https://github.com/chdb-io/chdb-node/issues/18

Also some feature implemented and bugs walked around before also need a better way to fix:

- #133 
- #176 
- #207 

Some chDB contributor also gave a try to make session better:

- https://github.com/chdb-io/chdb/pull/108

<div type='discussions-op-text'>

<sup>Originally posted by **l1t1** February  7, 2024</sup>
```sql
:) create table a engine=Memory as select 1 a;

0.11995077133178711
:) select * from a;
Code: 60. DB::Exception: Table _local.a does not exist. (UNKNOWN_TABLE)

0.11473512649536133
```

Here is how clickhouse-local interactive mode works:

```
root@0a8b55995b6e:/auxten/chdb/tests# ./ch24.5/usr/bin/clickhouse
ClickHouse local version 24.5.1.1763 (official build).

0a8b55995b6e :) create table a engine=Memory as select 1 a;

CREATE TABLE a
ENGINE = Memory
AS SELECT 1 AS a

Query id: 967a5d72-bb39-4a42-8a11-a108eda2a5d9

Ok.

0 rows in set. Elapsed: 0.008 sec. 

0a8b55995b6e :) select * from a;

SELECT *
FROM a

Query id: e5be6b9b-b752-4418-8753-adb8cc69a127

   ┌─a─┐
1. │ 1 │
   └───┘

1 row in set. Elapsed: 0.008 sec. 
```

</div>

The good part are:
1. Better support for states like 'Memory Table Engine', 'UDF', 'SET', 'USE'
2. Less tricky code to handle default database and 'SET', 'USE' statements
3. Without load tables and do init on every `query` function call, Performance should be much better than current implementation

auxten:
Fix #197

### DIFF
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -27,6 +27,8 @@ Checks: [
   '-bugprone-not-null-terminated-result',
   '-bugprone-reserved-identifier', # useful but too slow, TODO retry when https://reviews.llvm.org/rG1c282052624f9d0bd273bde0b47b30c96699c6c7 is merged
   '-bugprone-unchecked-optional-access',
+  '-bugprone-crtp-constructor-accessibility',
+  '-bugprone-suspicious-stringview-data-usage',
 
   '-cert-dcl16-c',
   '-cert-dcl37-c',
@@ -36,6 +38,7 @@ Checks: [
   '-cert-msc51-cpp',
   '-cert-oop54-cpp',
   '-cert-oop57-cpp',
+  '-cert-err33-c', # Misreports on clang-19: it warns about all functions containing 'remove' in the name, not only about the standard library.
 
   '-clang-analyzer-optin.performance.Padding',
 
@@ -94,10 +97,12 @@ Checks: [
   '-modernize-pass-by-value',
   '-modernize-return-braced-init-list',
   '-modernize-use-auto',
+  '-modernize-use-constraints', # This is a good check, but clang-tidy crashes, see https://github.com/llvm/llvm-project/issues/91872
   '-modernize-use-default-member-init',
   '-modernize-use-emplace',
   '-modernize-use-nodiscard',
   '-modernize-use-trailing-return-type',
+  '-modernize-use-designated-initializers',
 
   '-performance-enum-size',
   '-performance-inefficient-string-concatenation',
@@ -121,7 +126,8 @@ Checks: [
   '-readability-magic-numbers',
   '-readability-named-parameter',
   '-readability-redundant-declaration',
-  '-readability-redundant-inline-specifier',
+  '-readability-redundant-inline-specifier', # useful but incompatible with __attribute((always_inline))__ (aka. ALWAYS_INLINE, base/base/defines.h).
+        # ALWAYS_INLINE only has an effect if combined with `inline`: https://godbolt.org/z/Eefd74qdM
   '-readability-redundant-member-init', # Useful but triggers another problem. Imagine a struct S with multiple String members. Structs are often instantiated via designated
         # initializer S s{.s1 = [...], .s2 = [...], [...]}. In this case, compiler warning `missing-field-initializers` requires to specify all members which are not in-struct
         # initialized (example: s1 in struct S { String s1; String s2{};}; is not in-struct initialized, therefore it must be specified at instantiation time). As explicitly
@@ -132,12 +138,7 @@ Checks: [
   '-readability-uppercase-literal-suffix',
   '-readability-use-anyofallof',
 
-  '-zircon-*',
-
-  # This is a good check, but clang-tidy crashes, see https://github.com/llvm/llvm-project/issues/91872
-  '-modernize-use-constraints',
-  # https://github.com/abseil/abseil-cpp/issues/1667
-  '-clang-analyzer-optin.core.EnumCastOutOfRange'
+  '-zircon-*'
 ]
 
 WarningsAsErrors: '*'
--- a/.cursorignore
+++ b/.cursorignore
@@ -1,0 +1,1 @@
+contrib/
--- a/chdb/__init__.py
+++ b/chdb/__init__.py
@@ -85,8 +85,10 @@ sql = query
 PyReader = _chdb.PyReader
 
 from . import dbapi, session, udf, utils  # noqa: E402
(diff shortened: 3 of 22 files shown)


ruslandoga:
👋

Since the PR was renamed to SQLite-like, I wonder if there are plans to allow reading the result blocks similar to `sqlite3_column_*` or DuckDB's `duckdb_fetch_chunk` and `duckdb_data_chunk_get_vector`? Or would it stay a single binary blob for now?

auxten:
Current progress:
- [x] SQLite like API: common functions, common connection strings supported
- [x] Stateful Query with long live clickhouse engine instance bind with connection
- [x] Python DB-API reimplemented with new API. Both persist and memory engine supported See: [tests/test_dbapi.py](https://github.com/chdb-io/chdb/blob/0f1ae8cb319156b1c9dab517a0ac5c21179c0619/tests/test_dbapi.py), [tests/test_dbapi_persistence.py](https://github.com/chdb-io/chdb/blob/0f1ae8cb319156b1c9dab517a0ac5c21179c0619/tests/test_dbapi_persistence.py)
- [x] ClickHouse memory engine support 
  - [x] #262 
  - [x] https://github.com/chdb-io/chdb/issues/258#issuecomment-2310208994
- [x] Some performance improvement (~43%)
     
image


About the chdb.connect
```
chdb.connect()

    Create a connection to chDB backgroud server.
    Only one open connection is allowed per process. Use `close` to close the connection.
    If called with the same connection string, the same connection object will be returned.
    You can use the connection object to create cursor object. `cursor` method will return a cursor object.

    Args:
        connection_string (str, optional): Connection string. Defaults to ":memory:".
        Aslo support file path like:
          - ":memory:" (for in-memory database)
          - "test.db" (for relative path)
          - "file:test.db" (same as above)
          - "/path/to/test.db" (for absolute path)
          - "file:/path/to/test.db" (same as above)
          - "file:test.db?param1=value1&param2=value2" (for relative path with query params)
          - "///path/to/test.db?param1=value1&param2=value2" (for absolute path)

        Connection string args handling:
          Connection string can contain query params like "file:test.db?param1=value1&param2=value2"
          "param1=value1" will be passed to ClickHouse engine as start up args.

          For more details, see `clickhouse local --help --verbose`
          Some special args handling:
            - "mode=ro" would be "--readonly=1" for clickhouse (read-only mode)

    Returns:
        Connection: Connection object
```

Example of memory engine of SQLite like API:
```python
# If you want best perf, the chdb.connect() API set will satisfy you
from chdb import connect

conn = connect(":memory:")
cursor = conn.cursor()
# Create a table
cursor.execute(
    """
    CREATE TABLE users (
        id Int32,
        name String,
        scores Array(UInt8)
    ) ENGINE = Memory
    """
)

# Insert test data
cursor.execute(
    """
    INSERT INTO users VALUES
    (1, 'Alice', [95, 87, 92]),
    (2, 'Bob', [88, 85, 90]),
    (3, 'Charlie', [91, 89, 94])
    """
)

# Test fetchone
cursor.execute("SELECT * FROM users WHERE id = 1")
row = cursor.fetchone()
```

More API see chdb.dbapi like:
```python
import tempfile
import unittest
from chdb import dbapi

with tempfile.TemporaryDirectory() as tmpdirname:
    conn = dbapi.connect(tmpdirname)
    print(conn)
    cur = conn.cursor()
    # cur.execute("CREATE DATABASE IF NOT EXISTS test_db ENGINE = Atomic")
    # cur.execute("USE test_db")
    cur.execute(
        """
    CREATE TABLE rate (
        day Date,
        value Int64
    ) ENGINE = ReplacingMergeTree ORDER BY day"""
    )

    # Insert single value
    cur.execute("INSERT INTO rate VALUES (%s, %s)", ("2021-01-01", 24))
    # Insert multiple values
    cur.executemany(
        "INSERT INTO rate VALUES (%s, %s)",
        [("2021-01-02", 128), ("2021-01-03", 256)],
    )
    # Test executemany outside optimized INSERT/REPLACE path
    cur.executemany(
        "ALTER TABLE rate UPDATE value = %s WHERE day = %s",
        [(72, "2021-01-02"), (96, "2021-01-03")],
    )

    # Test fetchone
    cur.execute("SELECT value FROM rate ORDER BY day DESC LIMIT 2")
    row1 = cur.fetchone()
    self.assertEqual(row1, (96,))
    row2 = cur.fetchone()
    self.assertEqual(row2, (72,))
    row3 = cur.fetchone()
    self.assertIsNone(row3)

    # Test fetchmany
    cur.execute("SELECT value FROM rate ORDER BY day DESC")
    result_set1 = cur.fetchmany(2)
    self.assertEqual(result_set1, ((96,), (72,)))
    result_set2 = cur.fetchmany(1)
    self.assertEqual(result_set2, ((24,),))

    # Test fetchall
    cur.execute("SELECT value FROM rate ORDER BY day DESC")
    rows = cur.fetchall()
    self.assertEqual(rows, ((96,), (72,), (24,)))

    # Clean up
    cur.close()
    conn.close()
```

For more please refer to:
- [tests/test_conn_cursor.py](https://github.com/chdb-io/chdb/blob/0f1ae8cb319156b1c9dab517a0ac5c21179c0619/tests/test_conn_cursor.py)
- [tests/test_dbapi.py](https://github.com/chdb-io/chdb/blob/0f1ae8cb319156b1c9dab517a0ac5c21179c0619/tests/test_dbapi.py)
- [tests/test_dbapi_persistence.py](https://github.com/chdb-io/chdb/blob/0f1ae8cb319156b1c9dab517a0ac5c21179c0619/tests/test_dbapi_persistence.py)

PR #290, opened 2024-12-19
Refactor query execution to use connection object

auxten:
- Refactor query execution to use connection object
- Handle query errors by throwing exceptions and release GIL during query execution


auxten:
1. Due to some limitation, the could be only 1 chDB connection(Session) in one process.

PR #367, opened 2025-08-24
Fix memory tracker

wudidapaopao:

...

<details>
    <summary>CI Settings</summary>

**NOTE:** If your merge the PR with modified CI you **MUST KNOW** what you are doing
**NOTE:** Checked options will be applied if set before CI RunConfig/PrepareRunConfig step

#### Run these jobs only (required builds will be added automatically):
- [ ]  Integration Tests
- [ ]  Stateless tests
- [ ]  Stateful tests
- [ ]  Unit tests
- [ ]  Performance tests
- [ ]  All with aarch64
- [ ]  All with ASAN
- [ ]  All with TSAN
- [ ]  All with Analyzer
- [ ]  All with Azure
- [ ]  Add your option here

#### Deny these jobs:
- [ ]  Fast test
- [ ]  Integration Tests
- [ ]  Stateless tests
- [ ]  Stateful tests
- [ ]  Performance tests
- [ ]  All with ASAN
- [ ]  All with TSAN
- [ ]  All with MSAN
- [ ]  All with UBSAN
- [ ]  All with Coverage
- [ ]  All with Aarch64

#### Extra options:
- [ ]  do not test (only style check)
- [ ]  disable merge-commit (no merge from master before tests)
- [ ]  disable CI cache (job reuse)

#### Only specified batches in multi-batch jobs:
- [ ]  1
- [ ]  2
- [ ]  3
- [ ]  4

</details>


src/Common/MemoryWorker.cpp
```diff
@@ -236,7 +236,7 @@ MemoryWorker::MemoryWorker(uint64_t period_ms_, bool correct_tracker_, bool use_
             static constexpr uint64_t cgroups_memory_usage_tick_ms{50};
 
             const auto [cgroup_path, version] = getCgroupsPath();
-            LOG_INFO(
+            LOG_TRACE(
```
auxten, line 239: It is really necessary? and the later mod

wudidapaopao: The INFO-level logs would be printed directly to standard output. For example, when a user runs a Python script, they would first see a string of content like "Will create cgroup reader from..." before the query results are displayed. Therefore, the log level has been changed to TRACE.

PR #398, opened 2025-10-13
Use macos-14 arm and fix pipeline

auxten:
Use macos-14 arm avoid homebrew not work on macos-13


kafka1991:
https://github.com/actions/runner-images/issues/13045
use `macos-15-intel` for mac x86

Issue #436, opened 2025-12-07
Support ClickHouse's query parameters


niyue:
# Use Case

ClickHouse natively supports `Query Parameters` feature [1][2], which allow users to write generic SQL templates using typed placeholders. This avoids manual string interpolation, improves readability, and reduces the risk of SQL injection.

# Sample usage
```python
import chdb

df = chdb.query(
    "SELECT toDate({base_date:String}) + number AS date "
    "FROM numbers({total_days:UInt64}) "
    "LIMIT {items_per_page:UInt64}",
    "DataFrame",
    params={"base_date": "2025-01-01", "total_days": 10, "items_per_page": 2},
)
print(df)

# expected output:
#         date
# 0 2025-01-01
# 1 2025-01-02
```

This is standard ClickHouse syntax, but `chdb` does not yet support passing query parameters.

# Describe the solution you'd like
Enable query parameters in `chdb` by wiring through ClickHouse’s native `setQueryParameters` C++ API.
Expose this capability across `chdb`’s language bindings (Python, Node.js, etc.) so end users can easily supply a params dictionary when executing a query.

# Additional context
This would bring `chdb` to parity with `clickhouse client CLI` and `clickhouse-connect` Python library

* ClickHouse client CLI supports this feature [2], e.g.
```shell
clickhouse-client --param_message='hello' --query="SELECT {message: String}"
hello
```
* clickhouse-connect, the official ClickHouse Python client library, supports similar features in its `QueryContext` API [3]
```python
client.create_query_context(query='SELECT value1, value2 FROM data_table WHERE key = {k:Int32}',
                            parameters={'k': 2},
                            column_oriented=True)
result = client.query(context=qc)
```

# References
[1] ClickHouse SQL syntax: defining and using query parameters, https://clickhouse.com/docs/sql-reference/syntax#defining-and-using-query-parameters
[2] How to Use Query Parameters in ClickHouse, https://clickhouse.com/videos/how-to-use-query-parameters-in-clickhouse
[3] clickhouse-connect client library's support for query parameters, https://clickhouse.com/docs/integrations/language-clients/python/advanced-querying#querycontexts

niyue:
I’ve submitted PR #437 to address this issue for the Python binding.

cyrusmsk:
What are the benefits of this approach compared to handling parameter incorporation on the language side?
Using f-strings or t-strings in Python let's say

niyue:
@cyrusmsk 
[...]

PR #437, opened 2025-12-07
feat: add support for ClickHouse's query parameters in python binding

niyue:
### Description
This PR adds support for ClickHouse query parameters and addresses #436.
It enables users to execute parameterized queries without manual string interpolation, resulting in more reusable query templates and reducing the risk of SQL injection.

### Changelog category
- New Feature

### Changelog entry
Add support for ClickHouse query parameters, allowing generic queries with typed placeholders.

`Query parameters` is a native ClickHouse feature, and the official documentation already covers their usage in detail, for example:
* [ClickHouse SQL syntax: defining and using query parameters](https://clickhouse.com/docs/sql-reference/syntax#defining-and-using-query-parameters)
* [How to Use Query Parameters in ClickHouse](https://clickhouse.com/videos/how-to-use-query-parameters-in-clickhouse)

This PR enables `chdb` users to take advantage of this feature directly.
A brief introduction and examples have been added to both `README.md` and `README-zh.md`.

### Test
Several Python unit tests have been added to validate this functionality.

### DIFF
--- a/tests/test_delta_lake.py
+++ b/tests/test_delta_lake.py
@@ -2,8 +2,6 @@
 
 import unittest
 import sys
-import platform
-import subprocess
 import chdb
 from chdb import session
 from utils import is_musl_linux
--- a/tests/test_query_params.py
+++ b/tests/test_query_params.py
@@ -5,9 +5,17 @@ import pandas as pd
 import chdb
 from chdb import session
 from chdb.state import connect
+from utils import is_musl_linux
 
 
 class TestQueryParams(unittest.TestCase):
+    def assert_exception_message(self, exc, expected):
+        message = str(exc.exception)
+        if is_musl_linux():
+            self.assertIn("Caught an unknown exception", message)
+        else:
+            self.assertIn(expected, message)
+
     def test_connection_query_with_params(self):
         df = chdb.query(
             "SELECT toDate({base_date:String}) + number AS d "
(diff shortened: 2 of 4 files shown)


niyue:
`chdb` is still quite new to me, so please pardon me if I’ve overlooked anything obvious — I’m happy to make any adjustments needed.

chdb/session/state.py
```diff
@@ -138,9 +138,14 @@ def query(self, sql, fmt="CSV", udf_path=""):
                 - "JSONCompact" - Compact JSON format
                 - "Arrow" - Apache Arrow format
                 - "Parquet" - Parquet format
+                - "DataFrame" - Pandas DataFrame
+                - "ArrowTable" - PyArrow Table
```
niyue: This part is not directly related to the issue, but I noticed that some formats defined in chdb were not documented in the docstring, so I updated it as well.

wudidapaopao:
Hi @niyue, thank you so much for your PR. Currently, there are two categories of test failures:
One category consists of query failures caused by network issues, which can be temporarily ignored.
The other category is failures occurring on musl Linux. The cause is that, due to implementation-specific limitations of musl Linux at present, all exceptions will trigger the error: RuntimeError: Caught an unknown exception.
A quick workaround is to add a conditional check in the test cases: if the current environment is musl Linux (you can use the utility function `is_musl_linux` provided in the test cases), directly treat all exception messages as containing the string "`Caught an unknown exception`".

niyue:
@wudidapaopao

Thanks for the detailed information. I’ve updated the tests, but I don’t have a musl Linux environment to verify locally, so I’ll have to rely on the CI to validate the changes. Hopefully this fixes the CI issues.

niyue:
@wudidapaopao This PR is ready for review. Please let me know if there’s anything else I should revise. Thanks!

Issue #439, opened 2025-12-09
Support AI-assisted SQL generation via ClickHouse client integration


niyue:
# Use Case
Starting from ClickHouse 25.7, [ClickHouse Client](https://clickhouse.com/docs/interfaces/cli) and [clickhouse-local](https://clickhouse.com/docs/operations/utilities/clickhouse-local) include [AI-powered functionality](https://clickhouse.com/docs/interfaces/cli#ai-sql-generation) that converts natural language descriptions into SQL queries. This feature allows users to describe their data requirements in plain text, which the system then translates into corresponding SQL statements. [1]

chdb could support similar capability without too much effort. This functionality is particularly valuable during exploratory data analysis. In environments such as Jupyter notebooks, an AI model can quickly interpret the table schema, generate representative sample data, and streamline a variety of routine tasks.

# Describe the solution you'd like

The ClickHouse client already implements this capability in C++, and chdb could leverage the same code path and expose it through its language bindings, such as Python. This avoids code duplication, and any future improvements made by the ClickHouse team would automatically benefit chdb as well.

# References
[1] AI-powered SQL generation, https://clickhouse.com/docs/use-cases/AI/ai-powered-sql-generation

niyue:
I’ve submitted a PR (https://github.com/chdb-io/chdb/pull/440) that introduces this support. Any feedback would be greatly appreciated. Thanks.

wudidapaopao:
ClickHouse uses the `??` prefix to denote the AI-powered SQL generation feature. Since chDB operates in a Python environment, could we directly add two types of interfaces? One could be named `generate_sql`, which directly generates SQL via a prompt. The other could be named `ask` or `chat`, which generates SQL, executes it directly, and returns the results to the user (with `Pandas DataFrame` as the default return format).

niyue:
That sounds reasonable. I consider this option previously but was hesitate because it would introduce new API which requires other bindings to implement and I would like to keep the change scope small initially. 

I will follow your advice to revise the code later. Thanks.

PR #440, opened 2025-12-09
feat: Support AI-assisted SQL generation via ClickHouse client integration

niyue:
This PR is still experimental, so please feel free to reject it if it doesn’t align with chdb’s direction. I would appreciate any discussion or feedback, particularly regarding the items listed in the TODO section.

# Description

This PR adds AI-powered SQL generation support to chdb, enabling users to translate natural language descriptions into executable SQL queries, and addresses https://github.com/chdb-io/chdb/issues/439. 

Starting from ClickHouse 25.7, both the ClickHouse client and clickhouse-local include built-in AI capabilities for SQL generation. This contribution brings similar functionality to chdb, improving usability in interactive and exploratory workflows.

# Solution
* The ClickHouse client’s C++ implementation is reused without modification to handle prompting, tooling, and communication with AI providers. See `src/Client/AI` for more details.

* chdb introduces the same `??` prefix used by the ClickHouse client and clickhouse-local as the trigger for AI-assisted query generation. It reuses the existing `query` API to accept user prompts, with additional glue code to integrate with the ClickHouse client’s AI-related components.

* AI-generated output is returned in `RAW` format, allowing users to easily extract the generated SQL string and execute it separately.

# Changelog category
* New Feature

# Changelog entry

Add AI-powered SQL generation support, enabling natural-language-to-SQL translation through chdb’s language bindings.

# Documentation entry for user-facing changes
* Users can trigger AI-powered SQL generation by prefixing the input with `??`. For example, `conn.query("?? list all users order by id")` may generate a SQL statement such as `SELECT * FROM users ORDER BY id`.
* This PR does not include user-facing documentation yet, as the current approach, reusing the query API, may not be ideal (see the TODO section). Once the API design is finalized, additional documentation will be added.

# Test
* I use [openrouter](https://openrouter.ai) to test several different models (all free), including:
  * z-ai/glm-4.5-air:free
  * qwen/qwen3-coder:free
  * openai/gpt-oss-120b:free
  * google/gemma-3n-e4b (tested locally via LM studio)
  
A single Python unit test has been added to validate this functionality. It requires certain environment variables to be set for the AI endpoint; otherwise, the test will be skipped.

# TODO
There are a few clear areas that could be improved, including:
* I’m not certain whether the existing query API is the best interface for this feature. A dedicated API could return a plain string directly and might offer a cleaner user experience. However, adding a new API would increase complexity and require implementation across all language bindings. If we continue using the existing `query` API, it becomes difficult to pass AI-related configuration parameters (such as `temperature`), since there’s no natural place to expose the options defined in `src/Client/AI/AIConfiguration.h`. I’d prefer not to bloat the current `query` API, and I’d welcome guidance on the preferred approach.
* CI will need to be configured with the appropriate AI API endpoint parameters for this feature to work reliably.
* The final output heavily depends on the model being used, as different models vary significantly in both code-generation quality and tool-use capabilities. To benefit from this enhancement, a reasonably strong model is expected. In my testing, `google/gemma-3n-e4b` is the weakest and barely usable, while `qwen/qwen3-coder` shows noticeably poorer tool-use capability compared to `z-ai/glm-4.5-air`.

# References
[1] AI-powered SQL generation, https://clickhouse.com/docs/use-cases/AI/ai-powered-sql-generation

# Demo
Here is a working demo video, consuming OpenAI compatible API from openrouter, using `z-ai/glm-4.5-air` as the model.

https://github.com/user-attachments/assets/933465bc-136f-4e88-9ab0-a695e7895413

### DIFF
--- a/README-zh.md
+++ b/README-zh.md
@@ -240,7 +240,8 @@ sess.close()
 
 更多示例，请参见 [examples](examples) 和 [tests](tests)。
 
-## AI 辅助 SQL 生成
+<details>
+  <summary><h4>🧠 AI 辅助 SQL 生成</h4></summary>
 
 chDB 可以将自然语言提示转换为 SQL。通过连接/会话字符串配置 AI 客户端参数：
 
@@ -268,11 +269,12 @@ conn.query("INSERT INTO nums VALUES (1), (2), (3)")
 sql = conn.generate_sql("Select all rows from nums ordered by n desc")
 print(sql)  # 例如：SELECT * FROM nums ORDER BY n DESC
 
-# ask() = generate_sql() + query()，format 与 query 一致
+# ask()：一键生成并执行 SQL
+# `ask()` 会先调用 `generate_sql` 再执行 `query`，关键字参数会透传给 `query`。
 print(conn.ask("List the numbers table", format="Pretty"))
 ```
 
-`Session` 同样支持以上能力；`Session.ask()` 默认返回 `CSV`（与 `ask`/`query` 的默认格式一致），其他格式请显式指定：
+`Session` 同样支持以上能力；`Session.ask()` 会将关键字参数透传给 `Session.query`：
 
 ```python
 from chdb import session as chs
@@ -283,6 +285,8 @@ with chs.Session("file::memory:?ai_provider=openai") as sess:
     df = sess.ask("Show all users ordered by id", format="DataFrame")
     print(df)
 ```
+
+</details>
 
 ## 演示和示例
 
--- a/README.md
+++ b/README.md
@@ -473,7 +473,8 @@ For more examples, see [examples](examples) and [tests](tests).
 
 <br>
 
-## AI-assisted SQL generation
+<details>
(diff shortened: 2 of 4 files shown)


src/Client/AI/AISQLGenerator.cpp
auxten: This seems unnecessary

niyue:
@auxten @wudidapaopao
This PR is ready for review. I’d really appreciate it if you could take a look, thanks!

Here’s a brief summary of the current status:

* Two new APIs, `generate_sql` and `ask`, have been added as suggested. For the `ask` API, I mirror all `query` parameters so that it doesn't introduce additional maintenance burden over time.

* Regarding the newly added Python tests:
  * To run them in CI, we need to configure an API key for an AI provider, and potentially update the model/provider base URL as well. If these are not configured, the tests will be skipped.
  * The tests are not yet 100% stable due to the inherent randomness of LLM outputs, especially when using the free models I tested with. Using a more capable LLM would likely make the tests more stable.

PR #509, opened 2026-01-31
Support query progress reporting

niyue:

  - New Feature

  ### Changelog entry

  Enable ClickHouse client-style progress bar and progress table output in chdb (Python), configurable via connection options.

  - [X] Documentation is written 

  Suggested doc text (to add under chdb Python connection options):

* Motivation
    * Long-running queries in Python scripts benefit from interactive feedback. This adds ClickHouse client-style progress bar/table output to chdb when enabled via connection options.

*  Parameters
    * progress=tty|err|off|default: enable/disable progress bar output
    * progress-table=tty|err|off|default: enable/disable progress table output
    * enable-progress-table-toggle=1/0: (default disabled in chdb) allow toggling progress table with Space in interactive mode

*  Example use
```python
  import chdb
  conn = chdb.connect(":memory:?progress=tty")
  conn.query("SELECT sum(number) FROM numbers_mt(1e10) GROUP BY number % 10 SETTINGS max_threads=4")

  conn = chdb.connect(":memory:?progress-table=tty")
  conn.query("SELECT sum(number) FROM numbers_mt(1e10) GROUP BY number % 10 SETTINGS max_threads=4")

  # progress bar via chdb.query options
  chdb.query(
      "SELECT sum(number) FROM numbers_mt(1e10) GROUP BY number % 10 SETTINGS max_threads=4",
      options={"progress": "tty"}
  )
```

  ## Summary

  This PR enables ClickHouse client-style progress output in `chdb`, including both progress bar and progress table, when configured via connection options. It ensures embedded client behavior writes to stdout and sends the required progress/profile events.

  ## Testing
  - Manual verification:
      - `python -c "import chdb; conn = chdb.connect(':memory:?progress=tty'); conn.query('SELECT sum(number) FROM numbers_mt(1e10) GROUP BY number % 10 SETTINGS max_threads=4')"`

  ## Implementation notes

  - Wire progress options into ChdbClient configuration, initialize TTY buffer, and enable progress/profile events.
  - Keep progress table toggle disabled by default in chdb (can be enabled explicitly).

### DIFF
--- a/chdb/state/sqlitelike.py
+++ b/chdb/state/sqlitelike.py
@@ -84,6 +84,20 @@ class StreamingResult:
         self._exhausted = False
         self._supports_record_batch = supports_record_batch
         self._is_dataframe = is_dataframe
+        self._progress_callback = None
+        self._cleanup_progress_callback = None
+        self._progress_callback_cleaned = False
+
+    def _bind_progress_cleanup(self, progress_callback, cleanup_progress_callback):
+        self._progress_callback = progress_callback
+        self._cleanup_progress_callback = cleanup_progress_callback
+
+    def _cleanup_progress_callback_once(self):
+        if self._progress_callback_cleaned:
+            return
+        self._progress_callback_cleaned = True
+        if self._cleanup_progress_callback is not None:
+            self._cleanup_progress_callback(self._progress_callback)
 
     def fetch(self):
         """Fetch the next chunk of streaming results.
@@ -123,15 +137,18 @@ class StreamingResult:
                 result = self._conn.streaming_fetch_df(self._result)
                 if result is None or result.empty:
                     self._exhausted = True
+                    self._cleanup_progress_callback_once()
                     return None
             else:
                 result = self._conn.streaming_fetch_result(self._result)
                 if result is None or result.rows_read() == 0:
                     self._exhausted = True
+                    self._cleanup_progress_callback_once()
                     return None
             return self._result_func(result)
         except Exception as e:
             self._exhausted = True
+            self._cleanup_progress_callback_once()
             raise RuntimeError(f"Streaming query failed: {str(e)}") from e
 
     def __iter__(self):
@@ -206,6 +223,10 @@ class StreamingResult:
                 self._conn.streaming_cancel_query(self._result)
             except Exception as e:
                 raise RuntimeError(f"Failed to cancel streaming query: {str(e)}") from e
+            finally:
+                self._cleanup_progress_callback_once()
+        else:
+            self._cleanup_progress_callback_once()
 
     def record_batch(self, rows_per_batch: int = 1000000) -> pa.RecordBatchReader:
         """
@@ -637,15 +658,26 @@ class Connection:
         if lower_output_format in _arrow_format:
             format = "Arrow"
 
-        c_stream_result = self._conn.send_query(query, format, params=params or {})
+        progress_callback = self._setup_auto_progress_callback()
+        try:
+            c_stream_result = self._conn.send_query(query, format, params=params or {})
+        except Exception:
+            self._cleanup_auto_progress_callback(progress_callback)
+            raise
+
         is_dataframe = lower_output_format == "dataframe"
-        return StreamingResult(
+        stream_result = StreamingResult(
             c_stream_result,
             self._conn,
             result_func,
             supports_record_batch,
             is_dataframe,
         )
+        stream_result._bind_progress_cleanup(
+            progress_callback,
+            self._cleanup_auto_progress_callback,
+        )
+        return stream_result
 
     def __enter__(self):
         """Enter the context manager and return the connection.
--- a/tests/test_progress_callback.py
+++ b/tests/test_progress_callback.py
@@ -1,0 +1,88 @@
+#!python3
+
+import unittest
+import chdb
+
+
+class TestProgressCallback(unittest.TestCase):
+    def test_query_with_progress_tty(self):
+        conn = chdb.connect(":memory:?progress=tty")
+        try:
+            result = conn.query("SELECT count() FROM numbers(1000000)", "CSV")
+        finally:
+            conn.close()
+
+        self.assertEqual(result.data().strip(), "1000000")
+
+    def test_query_with_progress_err(self):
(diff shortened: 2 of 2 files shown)


chdb/__init__.py
```diff
@@ -167,27 +169,30 @@ def query(sql, output_format="CSV", path="", udf_path="", params=None):
 
         >>> # Query with UDF
         >>> result = chdb.query("SELECT my_udf('test')", udf_path="/path/to/udfs")
+
+        >>> # Query with progress bar
+        >>> result = chdb.query("SELECT 1", options={"progress": "tty"})
```
niyue: I would like to make API consistent, and besides creating `connection` with args approach, users could use this feature in `chdb.query` API as well, but I don't want to make `chdb.query` API bloated by introducing more keyword args, so I add a new generic `options` argument and all ClickHouse startup arguments could be passed in via this argument. Let me know if this is desirable. Thanks.

niyue:
Here are the latest screen casts as demo (please pay attention to the `progress=tty` and `progress-table=tty` options in the connection string:
# progress bar
progress_bar

# progress table
progress_table

niyue:
@auxten @wudidapaopao 
This PR is ready for review. I’d appreciate your feedback when you have a moment. Thanks!

auxten:
Can it work in jupyter notebook?
Also, what's the difference between `tty` and `err` (perhaps `stderr` is less ambiguous)?
Can we directly add `auto` to let chdb automatically choose for the user?

niyue:
> what's the difference between `tty` and `err` (perhaps `stderr` is less ambiguous)?
> Can we directly add auto to let chdb automatically choose for the user?

`tty` represents `/dev/tty`, and `err` represents `stderr`. `/dev/tty` is the current process's controlling terminal device file, it is used by ClickHouse client due to this:

> If /dev/tty is available, the progress in clickhouse-client and clickhouse-local will be rendered directly to the terminal, without writing to STDERR. It allows getting progress even if STDERR is redirected to a file, and the file will not be polluted by terminal escape sequences. (see [1])

In this pull request, the option value is passed directly to the ClickHouse client, which keeps the chdb implementation simpler. The ClickHouse client uses `err` instead of `stderr` and supports additional values such as `off`, `default`, `on`, `yes`, etc (see the critical code snippet in [3]).

Since the upstream ClickHouse code may evolve over time, keeping this option as a pass-through allows chdb to automatically benefit from upstream changes without introducing extra maintenance overhead.

The `auto` value seems to me similarly as `default` in ClickHouse client. If we want to further improve the detection of the default output destination, that enhancement would likely be better implemented in ClickHouse itself, so that chdb can benefit from it automatically in the future.

---
[1] ClickHouse release notes, progress bar support, https://clickhouse.com/docs/whats-new/changelog/2022#improvement-1
[2] ClickHouse pull request, write progress directly to terminal, https://github.com/ClickHouse/ClickHouse/pull/42003
[3] ClickHouse client `ClientBase.cpp`'s handling of progress related options, https://github.com/ClickHouse/ClickHouse/blob/master/src/Client/ClientBase.cpp#L229-L236

niyue:
> Can it work in jupyter notebook?

I tested both Jupyter and Marimo, and the current implementation doesn’t work properly in either environment. The ClickHouse client outputs progress information to `/dev/tty` or `stderr`, but neither approach works well in notebook environments.

However, I noticed that `tqdm` (the Python de facto progress library) works correctly in both Jupyter and Marimo, likely due to some notebook-specific UI handling. I took a brief look at `tqdm`, and it appears to provide specialized integrations such as `tqdm.notebook`, `tqdm.rich`, and `tqdm.gui`, which may be relevant here. I’m not yet familiar with its internal implementation, so I’m not sure whether a similar approach would be suitable for chdb, and I plan to investigate this further.

If it turns out to be feasible and a good fit, this could probably be implemented as a separate PR.

[1] tqdm.notebook, https://tqdm.github.io/docs/notebook/

niyue:
After some more investigation, now I implemented a Python-facing progress enhancement for notebook environments using a new progress callback API (like DuckDB's programmable progress callback).

### What changed
* Added a structured progress callback path from C++ to Python.
* Added notebook runtime detection (Jupyter/Marimo) in Python bindings.
* When `progress=auto` is used in notebook contexts, chDB now renders a notebook-friendly progress bar via callback updates.
* Kept terminal behavior unchanged for existing progress modes (progress=tty|err|off, progress-table=tty|err|off).
* A new `set_progress_callback` API is added to `Connection` so that users can register their own progress callback for rendering

### Implementation brief summary
* Progress accumulation is handled in `ClientBase`, so callback consumers receive accumulated values directly.
* Local/chdb-specific layer only forwards callback payloads to Python.

### Current scope
* Notebook rendering supports progress bar only. Progress table is not implemented for notebook rendering in this iteration so that we don't have to change too much upstream code (progress table requires a more complex callback API that returns profile events)

niyue:
I tested against Jupyter Lab 4.5.3 and Marimo 0.18.2, and here is a demo for the progress bar rendering in marimo notebook:

https://github.com/user-attachments/assets/4c340b30-f834-4f0f-b6a8-f34e16c99a14

niyue:
This is a minimum example for registering a user provided customized progress callback, which simply prints the progress to the stdout:
```python
import sys
import chdb

def progress_cb(p: dict) -> None:
    read_rows = int(p.get("read_rows", 0))
    total_rows = int(p.get("total_rows_to_read", 0))
    if total_rows > 0:
        pct = min(100.0, read_rows * 100.0 / total_rows)
        sys.stdout.write(f"\rProgress: {read_rows}/{total_rows} ({pct:5.1f}%)")
    else:
        sys.stdout.write(f"\rProgress: {read_rows} rows")
    sys.stdout.flush()

conn = chdb.connect(":memory:?progress=auto")
conn.set_progress_callback(progress_cb)
conn.query("SELECT sum(number) FROM numbers_mt(1e10) GROUP BY number % 10 SETTINGS max_threads=4")
print()
```

niyue:
@auxten 
I've updated the PR to add support for marimo and Jupyter via a DuckDB-like `set_progress_callback` API. This requires touching a few more upstream files, so the scope is slightly broader now.
Could you please take a look and let me know whether you think it's worth proceeding in this direction? Thanks!

Issue #519, opened 2026-02-11
Flaky Test: tests/test_json_accessor.py


wudidapaopao:
**Description**
Test name: tests/test_json_accessor.py::TestJsonExtractFloat::test_json_extract_float
Failure reason: AssertionError
CI report: https://github.com/chdb-io/chdb/actions/runs/21851758221/job/63063579672?pr=515

PR #520, opened 2026-02-12
Fix unstable test_json_accessor.py

wudidapaopao:
Fixed https://github.com/chdb-io/chdb/issues/519

- Build/Testing/Packaging Improvement

### Changelog entry (a [user-readable short description](https://github.com/ClickHouse/ClickHouse/blob/master/docs/changelog_entry_guidelines.md) of the changes that goes into CHANGELOG.md):
...

<details>
    <summary>CI Settings</summary>

**NOTE:** If your merge the PR with modified CI you **MUST KNOW** what you are doing
**NOTE:** Checked options will be applied if set before CI RunConfig/PrepareRunConfig step

#### Run these jobs only (required builds will be added automatically):
- [ ]  Integration Tests
- [ ]  Stateless tests
- [ ]  Stateful tests
- [ ]  Unit tests
- [ ]  Performance tests
- [ ]  All with aarch64
- [ ]  All with ASAN
- [ ]  All with TSAN
- [ ]  All with Analyzer
- [ ]  All with Azure
- [ ]  Add your option here

#### Deny these jobs:
- [ ]  Fast test
- [ ]  Integration Tests
- [ ]  Stateless tests
- [ ]  Stateful tests
- [ ]  Performance tests
- [ ]  All with ASAN
- [ ]  All with TSAN
- [ ]  All with MSAN
- [ ]  All with UBSAN
- [ ]  All with Coverage
- [ ]  All with Aarch64

#### Extra options:
- [ ]  do not test (only style check)
- [ ]  disable merge-commit (no merge from master before tests)
- [ ]  disable CI cache (job reuse)

#### Only specified batches in multi-batch jobs:
- [ ]  1
- [ ]  2
- [ ]  3
- [ ]  4

</details>

### DIFF
--- a/datastore/tests/test_json_accessor.py
+++ b/datastore/tests/test_json_accessor.py
@@ -161,7 +161,7 @@ class TestJsonExtractFloat:
     def test_json_extract_float(self, ds_float_json):
         """Test extracting float values from JSON."""
         ds_float_json['price'] = ds_float_json['data'].json.json_extract_float('price')
-        df = ds_float_json.to_df()
+        df = ds_float_json.to_df().sort_values('id')
 
         expected = [19.99, 29.50]
         assert list(df['price']) == pytest.approx(expected)
@@ -363,7 +363,7 @@ class TestJsonArrayExtraction:
         The fix detects this and falls back to JSON parsing in Python.
         """
         ds_json_arrays['tags_raw'] = ds_json_arrays['data'].json.json_extract_array_raw('tags')
-        df = ds_json_arrays.to_df()
+        df = ds_json_arrays.to_df().sort_values('id')
 
         # Result should be a list
         result = df['tags_raw'].iloc[0]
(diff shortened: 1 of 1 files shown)


auxten:
Best effort same behavior with Pandas including data types and order is our design.

This PR masks the symptom rather than fixing the root cause, which doesn't align with our testing principles:
1. Band-aid fix: Adding .to_df().sort_values('id') after execution hides a DataStore bug — the fixtures already have ORDER BY id, so if row order is non-deterministic, the issue is in DataStore's execution pipeline, not the tests.
1. Inconsistent with existing tests in the same file: Lines 72/83/94 already use the correct pattern ds.sort_values('id').to_df() (lazy, DataStore-level sort). This PR adds 7 instances of ds.to_df().sort_values('id') (pandas-level sort), which bypasses the lazy pipeline.

If a quick fix is needed, at minimum unify to ds.sort_values('id').to_df(). But the real fix should investigate why DataStore.from_df(df) loses row order from the source DataFrame.

PR #561, opened 2026-04-16
Add DataFrame writeback to remote ClickHouse server

### LINKED ISSUE #560: Support DataFrame writeback to remote ClickHouse server
Add writeback APIs (`save`, `to_clickhouse`, `create_view`, `create_materialized_view`) to write DataStore data back to a remote ClickHouse server.

wudidapaopao:
Closes #560

Add writeback APIs to DataStore for writing data back to a remote ClickHouse server.

## Changes

- `save()` — unified entry point
- `to_clickhouse()` — write data to a remote table (`fail` / `replace` / `append`)
- `create_view()` / `create_materialized_view()` — create views on remote server
- DataFrame upload fallback for Pandas-only pipelines

### DIFF
--- a/.github/ci/clickhouse-pair.yml
+++ b/.github/ci/clickhouse-pair.yml
@@ -1,0 +1,73 @@
+# Two independent ClickHouse instances used by the writeback integration tests.
+#
+# This file is the single source of truth for the test environment; CI and
+# local developers share it. Bring the pair up with:
+#
+#   docker compose -f .github/ci/clickhouse-pair.yml up -d --wait
+#
+# Tear down with:
+#
+#   docker compose -f .github/ci/clickhouse-pair.yml down -v
+#
+# Once running, point datastore/tests/test_writeback.py at it via:
+#
+#   CHDB_TEST_HOST=ch1:9000              CHDB_TEST_USER=remote_user
+#   CHDB_TEST_PASSWORD=test123           CHDB_TEST_DATABASE=chdb_test
+#   CHDB_TEST_TARGET_HOST=localhost:9001 CHDB_TEST_TARGET_USER=remote_user_2
+#   CHDB_TEST_TARGET_PASSWORD=test456    CHDB_TEST_TARGET_DATABASE=chdb_test_2
+#
+# Both nodes are wired to the same docker network, so a query running on `ch1`
+# can reach `ch2` via the hostname `ch2:9000` (and vice versa). This matches
+# the cross-server scenarios the test suite exercises.
+#
+# Why CHDB_TEST_HOST=ch1:9000 (not localhost:9000): cross-server writeback runs
+# `INSERT ... SELECT ... FROM remote(<source-host>, ...)` *on the target server*
+# (ch2). `localhost:9000` would resolve to ch2 itself there, so the source host
+# must be a name that points to ch1 from both ch2 (docker DNS resolves `ch1`)
+# and from your shell, where the test process + local chDB run. Add a hosts
+# alias so your shell resolves it to the published port too:
+#
+#   echo "127.0.0.1 ch1" | sudo tee -a /etc/hosts
+
+name: chdb-ds-writeback
+
+services:
+  ch1:
+    image: clickhouse/clickhouse-server:24.8
+    container_name: chdb_ds_ch1
+    # Map ch1's native port to the host so test code (running outside the
+    # docker network) can reach it as localhost:9000.
+    ports:
+      - "9000:9000"
+      - "8123:8123"
+    ulimits:
+      nofile:
+        soft: 262144
+        hard: 262144
+    volumes:
+      - ./init-ch1.sql:/docker-entrypoint-initdb.d/init.sql:ro
+    healthcheck:
+      test: ["CMD-SHELL", "wget -qO- http://localhost:8123/ping || exit 1"]
+      interval: 3s
+      timeout: 3s
+      retries: 30
+      start_period: 5s
+
+  ch2:
+    image: clickhouse/clickhouse-server:24.8
+    container_name: chdb_ds_ch2
+    ports:
+      - "9001:9000"
+      - "8124:8123"
+    ulimits:
+      nofile:
+        soft: 262144
+        hard: 262144
+    volumes:
+      - ./init-ch2.sql:/docker-entrypoint-initdb.d/init.sql:ro
+    healthcheck:
+      test: ["CMD-SHELL", "wget -qO- http://localhost:8123/ping || exit 1"]
+      interval: 3s
+      timeout: 3s
+      retries: 30
+      start_period: 5s
--- a/.github/ci/init-ch1.sql
+++ b/.github/ci/init-ch1.sql
@@ -1,0 +1,18 @@
+-- Source ClickHouse instance — referenced as `localhost:9000` from tests and
+-- as `ch1:9000` from inside the docker network.
+--
+-- Files in /docker-entrypoint-initdb.d/ are executed by the official image
+-- exactly once on first container startup, after the server is up. If you
+-- need to re-run them, recreate the volume (`docker compose down -v`).
+
+CREATE DATABASE IF NOT EXISTS chdb_test;
+
+CREATE USER IF NOT EXISTS remote_user
+    IDENTIFIED WITH plaintext_password BY 'test123'
+    HOST ANY;
+
+GRANT ALL ON chdb_test.* TO remote_user;
+
+-- The library probes existence via `SELECT count() FROM remote(..., 'system',
+-- 'tables', ...)`, so the test user must be able to read system metadata.
+GRANT SELECT ON system.* TO remote_user;
--- a/.github/ci/init-ch2.sql
+++ b/.github/ci/init-ch2.sql
@@ -1,0 +1,23 @@
+-- Target ClickHouse instance — referenced as `localhost:9001` from tests and
+-- as `ch2:9000` from inside the docker network.
+--
+-- Distinct credentials from ch1 on purpose: a misrouted connection (e.g. a
+-- code path that accidentally sends source creds to the target) gets
+-- rejected immediately rather than silently passing.
+
+CREATE DATABASE IF NOT EXISTS chdb_test_2;
+
+CREATE USER IF NOT EXISTS remote_user_2
+    IDENTIFIED WITH plaintext_password BY 'test456'
+    HOST ANY;
+
+GRANT ALL ON chdb_test_2.* TO remote_user_2;
+GRANT SELECT ON system.* TO remote_user_2;
+
+-- Cross-server writeback runs `INSERT INTO <target> SELECT ... FROM
+-- remote('ch1:9000', ...)` *on this (target) server*, so remote_user_2 must
+-- hold the global privileges the remote() table function requires. These are
+-- not covered by the database-scoped `GRANT ALL ON chdb_test_2.*` above:
+--   * REMOTE                 — permission to use remote()/remoteSecure()
+--   * CREATE TEMPORARY TABLE — remote() materializes its result set into one
+GRANT REMOTE, CREATE TEMPORARY TABLE ON *.* TO remote_user_2;
--- a/.github/workflows/external_clickhouse_integration.yaml
+++ b/.github/workflows/external_clickhouse_integration.yaml
@@ -1,0 +1,208 @@
+name: External ClickHouse Integration
+
+# Integration tests against a pair of real ClickHouse servers — currently
+# exercises the writeback APIs (to_clickhouse / create_view /
+# create_materialized_view / save) in datastore/tests/test_writeback.py.
+#
+# Only runs on Linux self-hosted runners (x86_64 + arm64). The macOS /
+# Windows wheel-build workflows do NOT run this suite — test_writeback.py's
+# autouse fixture pytest.skip()s when CHDB_TEST_* env vars are unset, so
+# those jobs see all writeback cases as SKIPPED at zero cost.
+#
+# The two ClickHouse servers are launched via docker compose
+# (.github/ci/clickhouse-pair.yml), which is also the file local devs use
+# to reproduce the CI environment.
+
+on:
+  workflow_dispatch:
+  push:
+    branches:
+      - main
+    paths-ignore:
+      - '**/*.md'
+  pull_request:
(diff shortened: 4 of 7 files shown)


wudidapaopao:
Thanks @Copilot — good catch, and it's worse than just `LIMIT`/`WHERE` ordering. Addressed in eba6e6f08de.

All server-side writeback SELECTs (`CREATE … AS SELECT`, `INSERT … SELECT`, VIEW/MV definitions, and the `DESCRIBE`-based schema inference) were built from the flat `to_sql()` / `_generate_select_sql()`, which ignores lazy-op ordering **and silently drops projections and groupby/agg (`LazySQLQuery`)**:

| pipeline | written before (buggy) | correct (execution_format) |
|---|---|---|
| `ds.head(5)[x > 0]` | `… WHERE x>0 LIMIT 5` (filter-then-limit) | `… (… LIMIT 5) WHERE x>0` |
| `ds.groupby('g').sum()` | `SELECT * FROM source` (**aggregation dropped — raw table written**) | full `GROUP BY` |
| `ds[['a','b']]` | `SELECT * FROM source` (**projection dropped**) | `SELECT a, b` |

This matches the intent already documented in `_is_fully_sql_pushable` ("the server-side path … uses `to_sql(execution_format=True)`").

Fix:
- New `_writeback_select_sql(same_server)` helper routes every writeback SELECT through `to_sql(execution_format=True)` — the exact SQL `_execute()` runs — so the three call sites can't drift.
- `_get_remote_view_sql()` now rebuilds on the execution-format SQL and folds `remote(source,…) → db.tbl` for both the aliased and the (execution-format) alias-less `FROM` forms.
- Schema inference (`_create_target_table` / `_apply_schema_evolution`) uses the same execution-format SELECT so CREATE/ALTER/INSERT agree.
- Pristine (no-lazy-op) pipelines are byte-identical to before.

Added regression tests: projection writes only the selected columns; `head()` before `filter()` preserves LIMIT-then-WHERE order.

wudidapaopao:
Thanks @Copilot — all three addressed in c170cfd4ae8:

1. **`_execute_remote_ddl` escaping**: `host`/`user`/`password` now go through the adapter's single-quote-doubling escaper (the same `_escape_sql_string` `_check_remote_table_exists` uses), so a quote in a connection param can't break out of the literal or inject SQL. The embedded query keeps backslash escaping (ClickHouse treats `\` as an escape inside string literals) **and** now doubles single quotes. The table-function name (`remote`/`remoteSecure`) is taken from the adapter so secure/cloud normalization is respected.
2. **`to_clickhouse(if_exists='replace')` docstring**: corrected — it's atomic `CREATE OR REPLACE TABLE … AS SELECT` for all SQL-pushable pipelines (same- or cross-server), and refuse-if-exists for the Pandas-only DataFrame path.
3. **`save(type='materialized_view')` docstring**: corrected — only `if_exists='fail'` is supported.

wudidapaopao:
Thanks @Copilot — all four addressed in 3bfc34c9401:

1. **`_target_server_context` credentials**: when `user`/`password` are omitted for a cross-server target, it now inherits the *source* connection's credentials (matching the `to_clickhouse`/`create_view`/`create_materialized_view` docstrings) instead of falling back to `default`/empty — which would have broken writeback to a target sharing non-default source creds. (Good catch — a real latent bug; the cross-server tests passed only because they pass explicit target creds.)
2. **`_materialize_for_writeback`**: replaced the `assert` with a real `ExecutionError` so the guard survives `python -O`.
3. **`_normalize_index_for_insert`**: raises `ValueError` when `index_label`'s length doesn't match the number of index levels (was silently truncating via `zip`).
4. **`to_clickhouse` docstring**: reworded the cross-server path — the CREATE/INSERT runs on the target server, which pulls source rows via `remote()`; chDB only ships the statement.

wudidapaopao:
Thanks @Copilot — all five addressed in 429786c75fa:

1. **`_execute_remote_ddl` normalization**: now takes `host`/`user`/`password` from the adapter (so ClickHouse-Cloud port/secure normalization is applied) instead of raw `_remote_params`, matching `_check_remote_table_exists`. Also fixed the stale "double quotes" comment — it doubles single quotes.
2. **`_target_server_context`**: `host` still defaults to the source server, so an explicit `user`/`password` now applies even when `host` is omitted (previously a `host=None` call silently dropped credential overrides); overriding nothing stays a no-op.
3. **`_parse_target_name` docstring**: corrected — it raises `QueryError` when no database can be inferred; it does not default to `"default"`.
4. **`create_view` docstring**: documented that a cross-server view persists the source credentials in its stored SQL (`remote(..., user, password)`), readable via `SHOW CREATE` / `system.tables` / the query log — prefer a least-privilege source user there.

wudidapaopao:
Thanks @Copilot.

- **Credential `None` edge case**: addressed in e04ad774cc4 — `_target_server_context` now treats a source `user`/`password` of `None` as missing (`or "default"` / `or ""`), so it can't flow an empty username into the adapter.

- **`sys.path` manipulation / "installs the wheel but imports the checkout"**: noted, but intentionally not changing it in this PR. The wheel is built from this same checkout (`make wheel`), so the `datastore` code under test is byte-identical, and `chdb-core` (the native engine) is installed from the wheel's dependencies and exercised. The `sys.path.insert(...)` is the repo-wide convention used by **every** test under `datastore/tests/` (not introduced by this PR); reworking import resolution risks the now-green suite for a narrow packaging-fidelity gain. Better handled as a separate, repo-wide change (e.g. a `conftest`/pytest importlib-mode switch) than a one-file divergence inside this feature PR.

All CI is green (Build & Test, both External ClickHouse with 60/60 writeback tests, all three DataStore platforms).

wudidapaopao:
Thanks @Copilot — all four addressed in e6bd29ccbcf:

1–3. **Backslash escaping**: `ClickHouseAdapter._escape_sql_string` now escapes backslashes (then doubles single quotes), since `\` is an escape character inside ClickHouse string literals. This single override fixes every call site at once — `_execute_remote_ddl`, `_check_remote_table_exists`, `_get_remote_table_schema`, `list_databases_sql`, etc. — and leaves the MySQL/PostgreSQL adapters on the base quote-only escaper (their string rules differ). `_execute_remote_ddl` now also reuses the adapter escaper for the embedded query (it's backslash-safe now).
4. **`_insert_dataframe_to_remote`**: runs the INSERT with `TabSeparated` output instead of the default `DataFrame` — the statement returns no rows.

wudidapaopao:
Thanks @Copilot — addressed in 7dd21098db1. Threaded an explicit `df` parameter through `_run_chdb_query` and `_describe_query_schema`, and pass the materialized DataFrame from `_create_target_table`, `_apply_schema_evolution`, and `_insert_dataframe_to_remote`. `Python(df)` now resolves in the exact frame that invokes `query()` — mirroring `connection.py:_execute_df_query` — instead of relying on chDB walking several stack frames up to find a `df` local. Verified locally that `DESCRIBE (SELECT * FROM Python(df))` resolves via the passthrough even when no parent frame has a variable named `df`.

wudidapaopao:
Thanks @Copilot — both addressed in 72d055b3769:

1. **`to_clickhouse` `order_by` docstring**: corrected — `order_by` isn't required for MergeTree; when omitted it defaults to `ORDER BY tuple()` (a key is recommended, not required). Kept the convenience default rather than enforcing.
2. **test `_ddl` helper**: now doubles single quotes (and escapes backslashes) for the embedded `remote(query=...)`, matching the production `_execute_remote_ddl` escaping — independent of any backslash-escape parser setting.

wudidapaopao:
Thanks @Copilot — addressed in 38b1dbab289. `_execute_remote_ddl` now catches `(RuntimeError, chdb.ChdbError)`: the module-level `chdb.query()` raises `RuntimeError`, but the persistent-connection path (`conn.query()`) can raise `chdb.ChdbError` (which is *not* a `RuntimeError` subclass), so a remote DDL/DML failure on a connected DataStore could previously leak a raw chdb exception. Both now normalize to `ExecutionError`.

wudidapaopao:
Thanks @Copilot — addressed in fa0a32c438b:

- **`to_clickhouse` / `save` Raises docstrings**: corrected — an existing target with `if_exists="fail"` is rejected by ClickHouse's CREATE and wrapped as `ExecutionError` (there's no library-side pre-check), not `QueryError`. `QueryError` still applies to the DataFrame-upload "replace" refusal and to MV validation. (Matches `test_fail_when_target_exists_raises`.)
- **`_target_server_context`**: the `host=None` no-op path now also applies the normalized effective credentials, so a source connection that stored `user`/`password` as `None` can't reach the adapter as an empty username.

On the **test-helper interpolation** (host/user/password/db/table in `_ddl` etc.): leaving as-is. Those values come from operator-controlled `CHDB_TEST_*` env vars with no quotes/backslashes, and the production paths are fully escaped now. Hardening the fixtures would touch many helpers for no real-world gain — better as a separate test-only cleanup if you'd like it.

wudidapaopao:
Thanks @Copilot — addressed in 0cb60b2e4ea:

- Added `_require_clickhouse_for_writeback()` to `to_clickhouse` / `create_view` / `create_materialized_view`: a non-ClickHouse source (`mysql`/`postgresql`/...) now fails early with a clear `QueryError` instead of emitting a `mysql()`/`postgresql()` table function that doesn't accept the `query=` form. (Pure-DataFrame stores remain caught earlier by the no-connection check, so the existing local-DataFrame test still raises `DataStoreError`.)
- Reworded the `_execute_remote_ddl` comment: `_escape_sql_string` escapes a value for embedding *inside* a single-quoted literal (the surrounding quotes are added separately), not a fully-quoted literal.

wudidapaopao:
Thanks @Copilot — you're right, addressed in 9e9b9ac791b. Moved backslash escaping into the base `SourceAdapter._escape_sql_string` (and dropped the now-redundant `ClickHouseAdapter` override). All adapters' SQL — `remote()`/`mysql()`/`postgresql()`/`mongodb()` table functions — is executed by chDB's ClickHouse engine, where backslash is special inside `'...'`, so every adapter now escapes backslashes as well as doubling single quotes.

wudidapaopao:
Thanks @Copilot:

- **secure without host**: addressed in ec3d8862dd5 — `_target_server_context` now applies `secure` in the `host=None` branch, so `to_clickhouse(secure=True)` / view / MV without an explicit host switches the source connection to TLS. Per your note, default behavior is unchanged: only an explicit `secure=True` upgrades (a `bool` can't distinguish omitted from explicit `False`, so it never silently downgrades a secure source).
- **`sys.path.insert(..., parent.parent)`**: leaving as-is. This is the repo-wide convention used across several `datastore/tests/*.py` files — it puts `datastore/` on the path so `from tests.test_utils import ...` resolves (`tests` as a top-level package); `import datastore` itself resolves via pytest's rootdir / the installed wheel. It's harmless here, and changing one file (or reworking it repo-wide) is outside this PR's scope.

wudidapaopao:
Thanks @Copilot:

- **`sudo -n` in the /etc/hosts step**: good catch — fixed in ab3d410ebbf (`sudo -n tee`, matching the pyenv build-deps step) so it fails fast rather than hanging on a password prompt.
- **`save(type=...)` shadowing the builtin**: keeping `type` — it's a deliberate, readable public-API name (`save("t", type="view")`), the shadow is local to the method (which never calls the `type()` builtin), and renaming would change the public signature.
- **`_build_engine_clause` case-sensitive MergeTree check**: intentional — ClickHouse table-engine names ARE case-sensitive (`ENGINE = mergetree()` is rejected with `Unknown table engine mergetree. Maybe you meant: ['MergeTree']`), so the case-sensitive `"MergeTree" in engine` check correctly mirrors ClickHouse; a lowercase engine fails clearly either way.

wudidapaopao:
Thanks @Copilot:

- **`assert local_df is not None`**: addressed in 38ab3fcfc72 — replaced with a real `ExecutionError` (consistent with the earlier `_materialize_for_writeback` change) so the internal invariant still holds under `python -O`.
- **test-helper escaping (`_ddl` / `_remote_func`)**: leaving as-is. These interpolate operator-controlled `CHDB_TEST_*` env values (host/user/password) that contain no quotes/backslashes; the production code paths are now fully escaped, and these are CI/dev fixtures, not a runtime security surface. I'm happy to do a dedicated test-fixtures escaping pass if maintainers want it, but it's out of scope for this feature PR.

wudidapaopao:
Thanks @Copilot:

- **`_require_clickhouse_for_writeback` too strict**: good catch — fixed in 43faf7f5e32. Promoted the set to a module constant `_CLICKHOUSE_SOURCE_TYPES = {"clickhouse", "remote", "remotesecure"}` (now also used by `__init__`'s `clickhouse_sources`), and the guard matches `source_type.lower()` against it. So `source="remote"`/`"remotesecure"` writeback is accepted again, while `mysql`/`postgresql` stay rejected.
- **test-helper escaping (`_ddl` / `_remote_func`)**: leaving as-is — test fixtures with operator-controlled `CHDB_TEST_*` values (no quotes/backslashes); production paths are fully escaped. (Same as noted in the prior rounds.)

wudidapaopao:
Thanks @Copilot — checked all three:

- **`_run_chdb_query` connection path returning a DataFrame**: this one is a false alarm. `_run_chdb_query` calls the *raw* `conn._conn.query(sql, "DataFrame")` (a `chdb.connect()` Connection) — not the streaming `query_df()` wrapper — and that raw call returns a pandas DataFrame. Verified empirically, and `connection.py` itself depends on exactly this (`self._conn = chdb.connect(...)`; `self._conn.query(sql, "DataFrame")` used as a DataFrame at lines 888/904/911/929/996, incl. `_execute_df_query`). So metadata/writeback work whether or not a persistent connection exists.
- **`_ddl` / `_remote_func` test-fixture escaping**: unchanged — operator-controlled `CHDB_TEST_*` values, production paths fully escaped.

I believe the genuine review items are now all addressed. Across the review rounds this covered: writeback SQL correctness (execution_format incl. the LazySQLQuery case), SQL-injection/backslash escaping across all adapters, credential + `secure` inheritance in the target-server context, exception normalization (`ChdbError`/asserts), the non-ClickHouse-source guard (incl. `remote`/`remotesecure` aliases), `Python(df)` frame binding, and many docstring corrections. Remaining comments are either not applicable (above) or intentional test-fixture hygiene. CI is green across all platforms.

ShawnChen-Sirius (approved):

PR #563, opened 2026-04-27
Fixes chdb-ds: UrlTableFunction.to_sql() emits HEADERS outside url() call

nklmish:
Per the [ClickHouse `url()` table function reference](https://clickhouse.com/docs/sql-reference/table-functions/url), headers must be the **4th positional argument** inside the function call, with structure as the 3rd positional argument (defaulting to `'auto'` when not provided — chdb rejects `''` with `Code: 36 Table structure is empty`).

## Changes

- Move headers into `sql_params` before `f"url(...)"` construction so they end up inside `url()` as the 4th positional arg
- Pad missing format/structure positional slots to `'auto'` via `sql_params.extend(["'auto'"] * (3 - len(sql_params)))`
- Convert `"Key: Value"` strings to `'Key'='Value'` pairs per ClickHouse's `headers()` function syntax
- Support `dict` input directly (`headers={"Authorization": "Bearer xyz"}`)
- Existing format/structure handling block is **preserved unchanged**

## Tests

New file `datastore/tests/test_url_table_function_headers.py` with `class TestUrlTableFunctionHeaders` (8 methods, matching the existing `test_table_functions.py` class-based convention):

- `test_to_sql_no_headers_unchanged` — regression guard for no-headers case
- `test_to_sql_with_headers_inside_url_call` — confirms headers emit inside `url()`
- `test_to_sql_headers_pad_structure_auto` — structure auto-padded when format provided
- `test_to_sql_headers_pad_format_and_structure_auto` — both auto-padded when neither provided
- `test_to_sql_headers_dict_input` — dict input support
- `test_to_sql_headers_multiple_entries` — multiple headers
- `test_to_sql_headers_missing_url_raises` — confirms the new headers code path doesn't suppress the existing url-required validation
- `test_to_sql_headers_executes_via_chdb` — end-to-end runtime check: the generated SQL parses and executes via `chdb.query()` against `https://httpbin.org/ip` (skips gracefully if chdb or network unavailable)

Updated existing `test_url_with_headers_list` and `test_url_with_headers_string` in `test_table_functions.py` — they were asserting the broken output, now assert the correct fixed shape.

## Verification

- All 10,688 existing `datastore/tests/` tests PASS (no regressions; 9 skipped, 85 xfailed, 4 xpassed all unchanged from baseline)
- 8 new tests PASS, including the runtime SQL execution check
- Empirically verified: fixed SQL executes cleanly against `https://httpbin.org/ip` with chdb 4.1.6 (Python 3.13, macOS arm64) — returns expected JSON, no parse error

## Issue

https://github.com/chdb-io/chdb/issues/564

### DIFF
--- a/.github/workflows/build_macos_arm64_wheels.yml
+++ b/.github/workflows/build_macos_arm64_wheels.yml
@@ -25,7 +25,7 @@ jobs:
     name: Test DataStore (macOS arm64)
     runs-on: macos-15-xlarge
     if: ${{ !github.event.pull_request.draft }}
-    timeout-minutes: 60
+    timeout-minutes: 120
     steps:
       - name: Check machine architecture
         run: |
--- a/datastore/table_functions.py
+++ b/datastore/table_functions.py
@@ -192,7 +192,15 @@ class UrlTableFunction(TableFunction):
         url: HTTP(S) URL to the data
         format: Input/output data format (optional, auto-detected if not provided)
         structure: Optional table structure
-        headers: Optional HTTP headers as list of 'Key: Value' strings
+        headers: Optional HTTP headers. Recommended: ``dict[str, str]``
+            (e.g. ``{"Authorization": "Bearer xxx"}``).
+            Legacy formats accepted for backward compatibility:
+
+            - ``list[str]`` of ``'Key: Value'`` strings
+            - a single ``'Key: Value'`` string
+
+            Generated as ClickHouse ``headers('Key'='Value', ...)`` and
+            placed as the 4th positional argument of ``url()``.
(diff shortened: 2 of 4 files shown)


wudidapaopao:
Hi @nklmish, thanks a lot for spotting this issue and providing the fix. Your changes are fully reasonable. I added escape handling, class comments and new test cases. Also fixed the edge case where format_name is empty but structure is set. Previously, the URL table function would ignore the structure parameter in this case.

nklmish:
Thank you @wudidapaopao for your kind support with adding new test cases and documentation !

PR #593, opened 2026-06-23
Add clickhouse-connect backend

ShawnChen-Sirius:
This PR makes chDB usable as a clickhouse-connect execution backend, so existing clickhouse-connect code runs unchanged against the embedded chDB engine instead of an HTTP server. It is the chDB-side half of the design that follows up on [ClickHouse/clickhouse-connect#753](https://github.com/ClickHouse/clickhouse-connect/pull/753); the matching clickhouse-connect change is [ClickHouse/clickhouse-connect#811](https://github.com/ClickHouse/clickhouse-connect/pull/811) and is required for this PR to function (it adds the entry-point mechanism this PR registers against).

## What this PR adds (6 commits, ~3,400 lines, all in this repo)

* `chdb/cc_backend.py` — `ChdbClient` is a thin `clickhouse_connect.driver.Client` subclass that overrides only the transport-shaped methods (`raw_query` / `raw_stream` / `raw_insert` / `command` / `_query_with_context` / `data_insert`) plus a true zero-copy Arrow fast path (`query_arrow` / `query_arrow_stream` / `query_df` ride chDB's in-process Arrow buffer through PyArrow's C Data Interface). `AsyncChdbClient` wraps the sync client via `run_in_executor`. `ChdbBackend` is the entry-point factory.
* `chdb/cc_extension.py` — the `client.chdb` namespace for chDB-only capabilities: `query_python(...)` (the `Python()` table function for in-process DataFrames), `register_function(...)` (Python UDFs), `cursor()` (chDB's native DB-API cursor), `session_path`. Absent on HTTP clients by design — accessing `.chdb` on an HTTP client raises `AttributeError`.
* `pyproject.toml` — registers chDB with clickhouse-connect via the `clickhouse_connect.backends` entry-point group, and adds clickhouse-connect as an optional dependency (`clickhouse-connect>=1.3.0`; will be bumped once the matching CC version is released).
* `tests/clickhouse_connect/test_cc_backend.py` — 105 tests relocated from clickhouse-connect PR #753, adapted to `backend=\"chdb\"`.
* `tests/clickhouse_connect/test_parity.py` — 38 chDB-vs-real-server output-parity assertions across primitives, dates, decimals, arrays, maps, tuples, enums, UUIDs, IPs, fixed strings, aggregates, group-by and datetime64.
* `scripts/cc_upstream_suite/` — harness that runs clickhouse-connect's own integration suite against `ChdbBackend` by redirecting `create_client` at import time. The skip handling is entirely data-driven (no `pytest.mark` capability markers — kept clickhouse-connect's repo minimal): `skip_list.txt` lists nodeid substrings for cases the embedded engine genuinely cannot satisfy (HTTP transport / RBAC / `external_data` / Arrow-vs-Native dtype divergence / etc.).
* `.github/workflows/clickhouse-connect-backend.yml` — three CI gates against the chDB backend: the relocated #753 suite (embedded), parity against a real ClickHouse 26.5 server (service container), and clickhouse-connect's own integration suite over a curated-green file set.

A few real-world fixes the integration-suite run forced:

* **Process-wide refcounted connection cache** in `cc_backend.py`. chdb-core's embedded engine permits only one active engine instance per process per data directory; a second `chdb.connect()` on the same path while the first is open deadlocks in the C layer. The cache lets many `ChdbClient` instances share one underlying connection.
* **`weakref.finalize` on every stream wrapper.** Guarantees the per-connection lock is released even when a test errors out without calling `close()`.
* **Lazy `USE` on `client.database = X`.** Over HTTP this is a per-request parameter and assigning a not-yet-created database is fine; chDB is session-scoped and needs `USE`, so we apply it lazily before the next operation, which preserves the common bootstrap of \"create database, then use it\".

## Testing

Empirical, no pytest markers:

| Gate | Result |
|---|---|
| Relocated backend suite (embedded) | 105 passed |
| Parity vs real ClickHouse 26.5.3.41 | 38 passed |
| clickhouse-connect's own integration suite vs chDB (with `skip_list.txt` applied) | **PASS=346, FAIL=0, ERROR=0, SKIP=181** across all 34 integration files |

Of the 423 cases passing on the HTTP baseline, 320 (76%) pass byte-identically on chDB; the remaining 103 are categorized in `skip_list.txt`:

| Category | Count | Reason |
|---|---:|---|
| external_data unsupported | 22 | engine lacks this capability |
| HTTP session_id / aiohttp pool / form-encoding / protocol version | 35 | embedded has no HTTP concepts |
| transport attributes (`headers` / compression / port) | 10 | embedded has no socket |
| RBAC (`GRANT` / `CREATE ROLE` / row policies) | 4 | engine has no access-control layer |
| `query_id` / per-query summary headers | 8 | not surfaced in-process |
| DateTime / Decimal / Enum / Time / Time64 formatting | 32 | Arrow-vs-Native representation differs (values correct) |
| column-name post-processing / server-formatted output | 12 | HTTP-side options not applicable |
| streaming / Arrow batch shape | 8 | chDB emits whole stream as one batch |
| pre-existing HTTP-baseline failures (not chDB-specific) | 34 | also fail on a real server in this provisioning |

`README.md` documents the version-subscription policy (personal fork branch → CC main → CC release) and the procedure to re-curate the skip list on each clickhouse-connect release.

## User demo

```python
# pip install 'clickhouse-connect[chdb]'
# (installs both; chdb self-registers via the `clickhouse_connect.backends` entry point)
import clickhouse_connect

# All downstream code is identical across backends.
client = clickhouse_connect.get_client(backend=\"chdb\", path=\":memory:\")

client.command(
    \"CREATE TABLE t (id UInt64, name String) ENGINE = MergeTree ORDER BY id\"
)

client.insert(\"t\", [[1, \"Alice\"]], column_names=[\"id\", \"name\"])

result = client.query(\"SELECT * FROM t\")
print(result.result_rows)
```

For chDB-only features:

```python
client.chdb.query_python(\"SELECT sum(a) FROM Python(my_df)\", my_df=my_df)
client.chdb.register_function(my_udf)
cursor = client.chdb.cursor()
```

From the user's point of view it is still the normal clickhouse-connect API; the chDB backend is selected with one keyword and chDB-only capabilities are reachable through an explicit, opt-in namespace.

## Release ordering

This PR depends on clickhouse-connect/#811 being merged and released. Once that's available, the `clickhouse-connect>=1.3.0` pin in `pyproject.toml` will be bumped to the version that ships the registry.

### DIFF
--- a/chdb/cc_backend.py
+++ b/chdb/cc_backend.py
@@ -5,7 +5,8 @@ This module makes chDB's in-process, embedded ClickHouse engine usable as a
 clickhouse-connect backend::
 
     import clickhouse_connect
-    client = clickhouse_connect.get_client(backend="chdb", path=":memory:")
+    client = clickhouse_connect.get_client("chdb://memory")               # in-memory
+    client = clickhouse_connect.get_client("chdb:///tmp/data.db")          # on-disk
     df = client.query_df("SELECT number FROM numbers(5)")
 
 It lives in the chdb-io/chdb repository -- *not* in clickhouse-connect -- and registers
@@ -1519,8 +1520,8 @@ class ChdbBackend:
     """Backend factory registered at the ``clickhouse_connect.backends`` entry point.
 
     Satisfies :class:`clickhouse_connect.driver.backend.Backend`: clickhouse-connect's
-    ``get_client(backend="chdb", ...)`` loads this object and calls ``create_client`` /
-    ``create_async_client``.
+    ``get_client("chdb://...")`` loads this object and calls ``create_client`` /
+    ``create_async_client`` with the URI's path and query parameters translated into kwargs.
     """
 
     backend_name = "chdb"
--- a/chdb/cc_backend.py
+++ b/chdb/cc_backend.py
@@ -1,0 +1,1488 @@
+"""
+chDB execution backend for clickhouse-connect.
+
+This module makes chDB's in-process, embedded ClickHouse engine usable as a
+clickhouse-connect backend::
+
+    import clickhouse_connect
+    client = clickhouse_connect.get_client(backend="chdb", path=":memory:")
+    df = client.query_df("SELECT number FROM numbers(5)")
+
+It lives in the chdb-io/chdb repository -- *not* in clickhouse-connect -- and registers
+itself through the ``clickhouse_connect.backends`` entry point (see ``pyproject.toml``).
+clickhouse-connect contains no chDB-specific code; the only coupling is the ``Backend``
+factory contract in ``clickhouse_connect.driver.backend`` and the entry-point name
+``chdb``. Everything here is owned and versioned by the chDB team, so chDB engine changes
+never require a clickhouse-connect change.
+
+``ChdbClient`` is a thin subclass of clickhouse-connect's ``Client``: it overrides only the
+transport-shaped methods (``raw_query`` / ``raw_stream`` / ``raw_insert`` / ``command`` /
+``_query_with_context``) plus the Arrow fast paths, and inherits the entire public API,
+type system, settings handling, Native parser, and error model unchanged. The same Native
+byte format the HTTP server emits is consumed verbatim, so result conversion is reused.
+
+Capability: ``supports_zero_copy_arrow = True``. ``query_arrow`` / ``query_arrow_stream``
+return PyArrow objects backed directly by chDB's in-process Arrow buffers (no wire bytes,
+no socket), and ``query_df`` routes through that zero-copy Arrow path. chDB-only features
+(the ``Python()`` table function, Python UDFs, the native DB-API cursor, the session path)
+are exposed through the ``client.chdb`` extension namespace (see ``cc_extension.py``), not
+bolted onto the base ``Client``.
+"""
+
+from __future__ import annotations
+
+import io
+import json
+import logging
+import os
+import re
+import sys
+import weakref
+import tempfile
+import threading
+from collections.abc import Generator, Iterable, Sequence
+from datetime import tzinfo
+from typing import TYPE_CHECKING, Any, BinaryIO
+
+from clickhouse_connect import common
+from clickhouse_connect.datatypes.registry import get_from_name
+from clickhouse_connect.driver.binding import bind_query, quote_identifier
+from clickhouse_connect.driver.client import Client
+from clickhouse_connect.driver.common import StreamContext, coerce_int
+from clickhouse_connect.driver.ctypes import RespBuffCls
+from clickhouse_connect.driver.exceptions import (
+    DatabaseError,
+    NotSupportedError,
+    ProgrammingError,
+    StreamFailureError,
+)
+from clickhouse_connect.driver.external import ExternalData
+from clickhouse_connect.driver.insert import InsertContext
+from clickhouse_connect.driver.options import check_arrow
+from clickhouse_connect.driver.query import QueryContext, QueryResult, TzMode, TzSource
+from clickhouse_connect.driver.summary import QuerySummary
+from clickhouse_connect.driver.transform import NativeTransform
+
+if TYPE_CHECKING:
+    import numpy
+    import pandas
+    import polars
+    import pyarrow
+
+logger = logging.getLogger(__name__)
+
+_columns_only_re = re.compile(r"LIMIT 0\s*$", re.IGNORECASE)
+
+# chdb's `send_query` emits each ClickHouse block as a self-contained encoding in the
+# requested format. For formats that have row-level (or block-level) self-description
+# and no global header/footer/file structure, concatenating chunks yields a valid
+# stream the caller's parser can consume directly. Other formats (Arrow, Parquet,
+# JSON, *WithNames variants, ...) would emit duplicated headers / multiple file
+# markers per chunk, which is not a valid larger stream. For those we fall back to a
+# single non-streaming query so the result is one well-formed payload.
+_STREAM_SAFE_FORMATS = frozenset(
+    {
+        "Native",
+        "TabSeparated",
+        "TSV",
+        "CSV",
+        "RowBinary",
+        "JSONEachRow",
+    }
+)
+
+
+def _finalize_stream(sr, lock, released_box):
+    """Idempotently close a chdb StreamingResult and release the per-connection lock.
+
+    Used both by an explicit ``close()`` call and by a ``weakref.finalize`` registered on
+    the stream wrapper, so a stream object that is GC'd without being closed (e.g. test
+    grabs the stream and then errors out) still releases the lock. Without this, a leaked
+    stream holds the shared connection lock forever and every subsequent query on any
+    ``ChdbClient`` using the same path deadlocks at ``with self._lock:``.
+
+    ``released_box`` is a single-element list shared with the wrapper instance so both
+    pathways agree on "already cleaned up".
+    """
+    if released_box[0]:
+        return
+    released_box[0] = True
+    try:
+        close = getattr(sr, "close", None)
+        if close is not None:
+            close()
+    except Exception:  # noqa: BLE001
+        logger.debug("Error closing chdb StreamingResult during finalize", exc_info=True)
+    try:
+        lock.release()
+    except RuntimeError:
+        pass
+
+
+class _BytesSource:
+    """
+    Minimal stand-in for the HTTP `ResponseSource` that the response buffer
+    expects. Yields a single chunk of bytes and exposes the attributes the
+    transform layer reads.
+    """
+
+    __slots__ = ("data", "last_message", "exception_tag")
+
+    def __init__(self, data: bytes):
+        self.data = data
+        self.last_message = None
+        self.exception_tag = None
+
+    @property
+    def gen(self):
+        def _gen():
+            yield self.data
+
+        return _gen()
+
+    def close(self):
+        return None
+
+
+class _ChdbStreamSource:
+    """
+    Source for `ResponseBuffer` backed by a chdb `StreamingResult`. Yields each
+    block's bytes and translates chdb's mid-stream RuntimeError into the
+    `StreamFailureError` clickhouse-connect callers expect.
+    """
+
+    __slots__ = ("_sr", "_released", "_finalizer", "last_message", "exception_tag", "__weakref__")
+
+    def __init__(self, streaming_result, lock: threading.Lock):
+        self._sr = streaming_result
+        self._released = [False]
+        # weakref.finalize guarantees lock release even if close() is never called (e.g. the
+        # stream is GC'd because a test errored out mid-iteration). See _finalize_stream.
+        self._finalizer = weakref.finalize(self, _finalize_stream, streaming_result, lock, self._released)
+        self.last_message = None
+        self.exception_tag = None
+
+    @property
+    def gen(self):
+        def _gen():
+            try:
+                while True:
+                    try:
+                        chunk = next(self._sr)
+                    except StopIteration:
+                        return
+                    except Exception as ex:  # noqa: BLE001
+                        raise StreamFailureError(_format_error_message(str(ex))) from ex
+                    payload = chunk.bytes() if hasattr(chunk, "bytes") else bytes(chunk)
+                    if payload:
+                        yield payload
+            finally:
+                self.close()
+
+        return _gen()
+
+    def close(self):
+        # Detach the finalizer and run the same cleanup directly (idempotent via the box).
+        self._finalizer()
+
+
+def _format_error_message(message: str) -> str:
+    """Extract a clean ClickHouse exception message from a chdb error string."""
+    if not message:
+        return ""
+    idx = message.find("Code: ")
+    if idx > 0:
+        return message[idx:].strip()
+    return message.strip()
+
+
+def _drain_to_bytes(block) -> bytes:
+    """Collect any supported insert_block shape into a single bytes value."""
+    if isinstance(block, (bytes, bytearray, memoryview)):
+        return bytes(block)
+    if isinstance(block, str):
+        return block.encode()
+    if hasattr(block, "to_pybytes"):
+        return block.to_pybytes()
+    if hasattr(block, "read"):
+        return block.read()
+    parts = []
+    for chunk in block:
+        parts.append(chunk if isinstance(chunk, (bytes, bytearray)) else chunk.encode())
+    return b"".join(parts)
+
+
+def _decompress(data: bytes, encoding: str) -> bytes:
+    if encoding == "lz4":
+        import lz4.frame
+
+        return lz4.frame.decompress(data)
+    if encoding == "zstd":
+        import zstandard
+
+        return zstandard.ZstdDecompressor().decompress(data)
+    if encoding == "gzip":
+        import gzip
+
+        return gzip.decompress(data)
+    if encoding == "br":
+        try:
+            import brotli
+        except ImportError as ex:
+            raise NotSupportedError("brotli is required to decompress 'br' for chdb raw_insert") from ex
+        return brotli.decompress(data)
+    if encoding == "deflate":
+        import zlib
+
+        return zlib.decompress(data)
+    raise NotSupportedError(f"Unsupported compression {encoding!r} for chdb raw_insert")
+
+
+# Process-wide cache of chdb connections, keyed by connection-string.
+#
+# chdb-core's embedded engine permits only one active engine instance per process per data
+# directory: a second ``chdb.connect("...")`` while a prior connection on the same path is
+# still open deadlocks inside the C extension. Because clickhouse-connect's tests freely
+# instantiate multiple clients (sync + async fixtures, parametrized factories, ...), the
+# backend must respect this and refcount-share a single underlying connection per path.
+# ``close()`` decrements; the chdb connection itself is only closed when the refcount drops
+# to zero. The lock guards the cache itself, not individual queries — each ChdbClient still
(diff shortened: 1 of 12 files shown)


ShawnChen-Sirius:
@chibugai review it

chdb/cc_extension.py
```diff
@@ -0,0 +1,119 @@
...
+
+        :param sql: SQL referencing the frames through ``Python(<name>)``.
+        :param fmt: ``"DataFrame"`` (default) returns a pandas DataFrame via the zero-copy
+            Arrow path; any other value is passed to chDB as the output format and the raw
+            result is returned.
+        :param frames: name -> Python object bindings visible to the ``Python()`` table function.
+        """
+        conn = self._client.chdb_connection
+        module_globals = globals()
+        with _python_tablefunc_lock:
```
chibugai, line 83: Thread-safety: this calls `conn.query` while holding only the extension-level `_python_tablefunc_lock`, but not the per-connection `self._client._lock`. cc_backend.py shares one chdb connection across all ChdbClient instances and serializes every `conn.query`/`send_query` via `_lock`, with a documented warning that a second concurrent query on the same connection deadlocks inside the engine. `_python_tablefunc_lock` only serializes globals publishing between `query_python` calls; it does not exclude concurrent regular queries (query/query_arrow/streaming) on the same connection, so a deadlock or corrupted result is possible. Acquire `self._client._lock` around the `conn.query` calls.

chdb/cc_backend.py
```diff
@@ -0,0 +1,1488 @@
...
+        self._chdb_path = chdb_path or ":memory:"
+        self._chdb_options = dict(chdb_options) if chdb_options else {}
+        self._connection_string = _build_conn_string(self._chdb_path, self._chdb_options)
+        self._chdb_module = chdb
+        # chdb-core permits only one open connection per data directory per process; reuse a
+        # refcounted shared connection across all ChdbClient instances for the same path.
+        # The shared per-connection lock serializes operations on it (a second concurrent
+        # chdb.send_query on the same connection deadlocks inside the engine).
+        self._conn, self._lock = _acquire_chdb_connection(self._connection_string)
+        self._closed = False
```
chibugai, line 368: Refcount leak risk: `_acquire_chdb_connection` increments the shared-connection refcount here, but if any subsequent step in `__init__` raises (`super().__init__`, `set_client_setting` loop, or the `ChdbExtension` import), `close()` is never called and the refcount stays elevated. The shared chdb connection for this path can then never be closed, and on `:memory:` paths it leaks indefinitely. Consider wrapping the remainder of `__init__` in try/except that calls `_release_chdb_connection(self._connection_string)` before re-raising.

chdb/cc_backend.py
```diff
@@ -0,0 +1,1488 @@
...
+            return sql + sep + extras.encode()
+        if " SETTINGS " in sql.upper():
+            return f"{sql}, {extras}"
+        return f"{sql} SETTINGS {extras}"
+
+    def _persist_setting(self, key: str, value: str) -> None:
+        """Apply a setting to the underlying chdb session via SET."""
+        try:
+            with self._lock:
+                self._conn.query(f"SET {key} = {self._quote_setting_value(value)}", "TabSeparated")
```
chibugai: The setting key is interpolated directly into the SQL while only the value is quoted via `_quote_setting_value`. If `key` reaches here unvalidated (e.g. arbitrary per-call/initial setting names), it can break the statement or inject SQL. `_validate_setting` only validates against known settings when `invalid_setting_action` is strict; with a permissive action an arbitrary key flows through `set_client_setting` -> `_persist_setting`. Consider validating the key against an allowed identifier pattern before interpolation.

chdb/cc_backend.py
```diff
@@ -0,0 +1,1488 @@
...
+                    for chunk in insert_block:
+                        tmp.write(chunk if isinstance(chunk, (bytes, bytearray)) else chunk.encode())
+            finally:
+                tmp.close()
+
+            per_call = self._filter_per_call_settings(settings)
+            settings_clause = (
+                f" SETTINGS {', '.join(f'{k} = {self._quote_setting_value(v)}' for k, v in per_call.items())}" if per_call else ""
+            )
+            sql = f"INSERT INTO {table}{cols} FROM INFILE '{tmp.name}'{settings_clause} FORMAT {fmt}"
```
chibugai: The temp file path is interpolated into a single-quoted INFILE literal without escaping. `tempfile` honors `TMPDIR`, and on systems where the temp directory path contains a single quote (or other SQL-special characters), this produces malformed SQL or breaks out of the literal. Escape `tmp.name` the same way `_quote_setting_value` escapes (replace `\` and `'`), or use the existing quoting helper.

scripts/cc_upstream_suite/_force_chdb_conftest.py
```diff
@@ -0,0 +1,103 @@
...
+# them, so strip before delegating (passing host=... to the chdb factory would be noise).
+_HTTP_ONLY = {
+    "host", "port", "interface", "secure", "dsn", "user", "username", "password",
+    "access_token", "token_provider", "compress", "client_name", "pool_mgr", "http_proxy",
+    "https_proxy", "verify", "ca_cert", "client_cert", "client_cert_key", "server_host_name",
+    "connect_timeout", "send_receive_timeout", "headers", "connector_limit",
+    "connector_limit_per_host", "keepalive_timeout",
+}
+
+_CHDB_PATH = os.getenv("CHDB_UPSTREAM_SUITE_PATH", ":memory:")
```
chibugai, line 47: Cross-test isolation risk: the chDB backend (`cc_backend.py`) shares a single process-wide, refcounted connection keyed by connection-string. With `_CHDB_PATH` defaulting to `:memory:` (and `run_upstream_suite.py` never setting `CHDB_UPSTREAM_SUITE_PATH`), every redirected `create_client()` in the upstream suite shares one `:memory:` engine. Tables/databases created by one test therefore persist into later tests, whereas the upstream HTTP suite assumes a fresh per-test server state. This can produce order-dependent / flaky failures that are hard to diagnose and may force unnecessary skip-list entries. Consider giving each test (or at least each client) an isolated path/database — e.g. derive a unique on-disk temp dir per worker/test, or DROP/CREATE a fresh database in a fixture — rather than a single shared `:memory:` instance.

scripts/cc_upstream_suite/run_upstream_suite.py
```diff
@@ -0,0 +1,105 @@
...
+        tmp = Path(tempfile.mkdtemp(prefix="cc-upstream-"))
+        repo = tmp / "clickhouse-connect"
+        shutil.copytree(args.cc_repo, repo, ignore=shutil.ignore_patterns(".git", "build", "*.egg-info", "__pycache__"))
+    else:
+        tmp = Path(tempfile.mkdtemp(prefix="cc-upstream-"))
+        repo = tmp / "clickhouse-connect"
+        _clone_tag(_installed_cc_version(), repo)
+
+    # Inject the redirect conftest at the checkout root (loaded before the integration conftest).
+    shutil.copyfile(INJECTED_CONFTEST, repo / "conftest.py")
```
chibugai: `copyfile` silently overwrites any existing root conftest.py in the upstream checkout. If clickhouse-connect ships a root conftest with required fixtures/hooks, replacing it could break collection and produce misleading gate results. Consider asserting it does not pre-exist (or chaining it) so silent loss of upstream setup is detected.

scripts/cc_upstream_suite/run_upstream_suite.py
```diff
@@ -0,0 +1,105 @@
...
+        sys.executable, "-m", "pytest", *args.tests.split(),
+        "-p", "no:cacheprovider", "-p", "no:xdist", "-o", "addopts=",
+        "-rsxX", *args.pytest_args,
+    ]
+    print(f"Running: {' '.join(cmd)}\n  cwd={repo}\n  skip_list={SKIP_FILE.name} xfail={XFAIL_FILE.name}")
+    rc = subprocess.run(cmd, cwd=repo, env=env)
+
+    if not args.keep and tmp is not None:
+        shutil.rmtree(tmp, ignore_errors=True)
+    return rc.returncode
```
chibugai: `tmp` is created via `mkdtemp` but cleanup only runs at the end of `main()`. If `_clone_tag`, `copytree`, `copyfile`, or the pytest subprocess raises, the temp directory leaks. Wrap the work in try/finally so cleanup runs on all paths (respecting `--keep`).

tests/clickhouse_connect/test_cc_backend.py
```diff
@@ -0,0 +1,1041 @@
...
+    assert r.result_rows == [(13, "user_1"), (79, "user_2")]
+
+
+@pytest.mark.parametrize("compression", ["lz4", "zstd", "gzip"])
+def test_raw_insert_decompresses_pre_compressed_payload(client, compression):
+    """raw_insert with `compression=<enc>` accepts compressed bytes and decompresses client-side."""
+    import gzip
+
+    import lz4.frame
+    import zstandard
```
chibugai: These optional compression libs are imported unconditionally, unlike `pandas`/`pyarrow`/`numpy` elsewhere which use `pytest.importorskip`. If `lz4` or `zstandard` is absent the whole parametrized test errors with ImportError at call time instead of skipping gracefully. Guard them, e.g. `lz4 = pytest.importorskip("lz4.frame")` / `zstandard = pytest.importorskip("zstandard")` before use.

tests/clickhouse_connect/test_cc_backend.py
```diff
@@ -0,0 +1,1041 @@
...
+    assert r.result_rows == [(13, "first"), (79, "second")]
+
+
+# ---- database parameter ----
+
+
+def test_database_parameter_switches_default():
+    c = clickhouse_connect.get_client(backend="chdb")
+    try:
+        c.command("CREATE DATABASE other_db")
```
chibugai, line 489: These tests create databases (`other_db`, `scoped_test`) on default `:memory:` clients, but the backend's `_CONN_CACHE` (cc_backend.py) keys the chdb connection by connection-string, so every default `:memory:` client shares one underlying connection refcounted process-wide. The comment's assumption that `:memory:` is per-connection isolated only holds once the refcount drops to zero between tests. If any other client (e.g. an overlapping fixture or a not-yet-GC'd client) keeps the refcount alive, the repeated `CREATE DATABASE other_db` / `scoped_test` will hit "database already exists" and make this test order-dependent/flaky. Consider using a unique `chdb_path=tmp_path/...` (as the persistence tests do) or unique DB names to guarantee isolation.

tests/clickhouse_connect/conftest.py
```diff
@@ -0,0 +1,31 @@
...
+"""
+
+import os
+import time
+from datetime import timezone
+
+import pytest
+
+os.environ["TZ"] = "UTC"
+time.tzset()
```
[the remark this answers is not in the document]
ShawnChen-Sirius: Resolved in f3d94f7. Done — guarded with hasattr(time, "tzset") so the module imports cleanly on Windows (the chDB backend itself raises NotSupportedError on win32 anyway).

chdb/cc_backend.py
```diff
@@ -0,0 +1,1488 @@
...
+        transport_settings: dict[str, str] | None = None,
+    ) -> bytes:
+        if external_data is not None:
+            raise NotSupportedError("external_data is not supported by the chdb backend")
+        final_query, bound = bind_query(query, parameters, self.server_tz)
+        # chdb's conn.query accepts both str and bytes; preserve bytes when binary
+        # parameter substitution (e.g. `$xx$` placeholders) yields non-UTF-8 SQL.
+        final_query = self._append_settings_clause(final_query, self._filter_per_call_settings(settings))
+        # HTTP path defaults to server's TabSeparated when no fmt is provided.
+        return self._exec_raw_query(final_query, fmt or "TabSeparated", params=self._strip_param_prefix(bound))
```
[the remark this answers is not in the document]
ShawnChen-Sirius: Resolved in f3d94f7. Done — raw_query now threads use_database into _exec_raw_query, which gates the lazy USE call on that flag.

chdb/cc_backend.py
```diff
@@ -0,0 +1,1488 @@
...
+        self._apply_database()
+        # Acquire the lock for the lifetime of the streaming read so concurrent
+        # callers don't interleave queries on the same chdb connection.
+        self._lock.acquire()
+        try:
+            streaming = self._conn.send_query(final_query, output_fmt, params=params or {})
+        except Exception as ex:  # noqa: BLE001
+            self._lock.release()
+            raise self._wrap_exception(ex) from ex
+        return _ChdbStreamFile(streaming, self._lock)
```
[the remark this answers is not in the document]
ShawnChen-Sirius: Resolved in f3d94f7. Done — same fix applied to raw_stream's streaming and non-streaming branches.

chdb/cc_backend.py
```diff
@@ -0,0 +1,1488 @@
...
+        while len(self._buf) < size and not self._eof:
+            chunk = self._pull()
+            if not chunk:
+                break
+            self._buf += chunk
+        if not self._buf:
+            return b""
+        out = self._buf[:size]
+        self._buf = self._buf[size:]
+        return out
```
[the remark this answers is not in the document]
ShawnChen-Sirius: Resolved in f3d94f7. Done — _ChdbStreamFile now uses a bytearray accumulator with extend / del-slice instead of bytes += chunk (amortized O(n) instead of O(n^2)).

tests/clickhouse_connect/test_cc_backend.py
```diff
@@ -0,0 +1,1041 @@
...
+
+# ---- query / command ----
+
+
+def test_command_returns_scalar(client):
+    assert client.command("SELECT 13") == 13
+    assert client.command("SELECT 'user_1'") == "user_1"
+
+
+def test_command_returns_tuple_for_multiple_columns(client):
```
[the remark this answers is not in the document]
ShawnChen-Sirius: Resolved in f3d94f7. Looking at this again — the assertion in this test is `assert t == (1, "a", 3.14)` which IS a tuple comparison, and the chDB backend's Native parser does return a tuple here (the test passes in our 105-case suite). I think Copilot misread the literal. Keeping the name and the assertion as-is; if you want it renamed for any reason let me know.

chdb/cc_extension.py
ShawnChen-Sirius: Resolved in f3d94f7. Done — query_python now acquires the per-connection lock (self._client._lock) around the conn.query call, so a concurrent regular query on the shared chdb connection cannot race with it.

chdb/cc_backend.py
ShawnChen-Sirius: Resolved in f3d94f7. Done — wrapped the remainder of __init__ in a try/except that calls _release_chdb_connection on any exception before re-raising, so a failure in super().__init__ / set_client_setting / extension import cannot leak the refcount.

chdb/cc_backend.py
ShawnChen-Sirius: Resolved in f3d94f7. Done — added _validate_setting_name (must match `^[A-Za-z_][A-Za-z0-9_]*$`) and applied it at every SET / SETTINGS interpolation site.

chdb/cc_backend.py
ShawnChen-Sirius: Resolved in f3d94f7. Done — added a module-level _quote_sql_string helper (same escape rules as _quote_setting_value) and used it in both INSERT INTO ... FROM INFILE sites in raw_insert and _insert_via_infile.

scripts/cc_upstream_suite/_force_chdb_conftest.py
ShawnChen-Sirius: Resolved in f3d94f7. Made this explicit: the runner now defaults CHDB_UPSTREAM_SUITE_PATH to a tempdir-per-invocation, so consecutive runs do not inherit on-disk state. Within a single run the shared :memory: behavior is intentional and matches HTTP — clickhouse-connect's own session fixture creates a random-named test database that all tests in the session share. Expanded the comment on _CHDB_PATH to document this trade-off explicitly.

scripts/cc_upstream_suite/run_upstream_suite.py
ShawnChen-Sirius: Resolved in f3d94f7. Done — replaced the silent copyfile with a pre-existence check that raises a clear SystemExit if a root conftest.py is already present in the checkout, so silent loss of upstream setup is impossible.

scripts/cc_upstream_suite/run_upstream_suite.py
ShawnChen-Sirius: Resolved in f3d94f7. Done — wrapped the work in try/finally so the temp checkout is cleaned on every path (respecting --keep).

tests/clickhouse_connect/test_cc_backend.py
ShawnChen-Sirius: Resolved in f3d94f7. Done — guarded both with pytest.importorskip("lz4.frame") / pytest.importorskip("zstandard") inside the parametrized test so a missing optional skips that case gracefully.

tests/clickhouse_connect/test_cc_backend.py
ShawnChen-Sirius: Resolved in f3d94f7. Done — switched test_database_parameter_switches_default to use a tmp_path on-disk chdb_path, so the test no longer races with the shared :memory: connection cache. The test_database_param_forwarded_to_use sibling was already using tmp_path.

wudidapaopao (approved):

PR #594, opened 2026-07-01
Add chdb.agents.ChDBTool — canonical agent tool + cross-language contract

ShawnChen-Sirius:
## What

Adds `chdb.agents.ChDBTool` — the canonical Python agent tool for chDB (query + introspection) — **plus a language-neutral contract and a shared conformance fixture** so chdb-io's per-language agent tools stay consistent instead of each re-implementing the surface differently.

This is a Python-level feature in the `chdb` package (alongside `datastore` / `cc_backend`), built on the `chdb-core` engine; it adds no C++/engine changes.

```python
from chdb.agents import ChDBTool

tool = ChDBTool(read_only=True)                       # SET readonly=2 on the session
tool.query("SELECT * FROM file('data.parquet') WHERE user_id = {u:String}",
           params={"u": user_id})                      # values bound, never concatenated
tool.describe("file('data.parquet')")                  # table OR table-function
tool.call("run_select_query", {"sql": "SELECT ..."})   # error-envelope path for agents
```

## Why

Multiple chdb-io surfaces already expose the same query/introspection shape independently — `mcp-clickhouse` (`run_chdb_select_query` + `list_databases`/`describe_table`/…), `chdb-node` (`chdbTools()`), and each future framework integration. Without a shared definition they drift: different read-only models, different error handling, different result caps. This PR makes the Python surface the reference implementation of **one contract** that every language conforms to, so:

- Python framework integrations (CrewAI, Pydantic AI, AutoGen, smolagents, LangChain query tool) become ~10-30 LOC shims over `ChDBTool` instead of full re-implementations.
- The TypeScript binding runs the **same** conformance fixture, so Python and TS behave identically.

## The contract (four pillars)

`chdb/agents/CONTRACT.md` is the source of truth; `chdb/agents/conformance/cases.jsonl` is its executable form.

1. **Read-only by the engine** — `SET readonly=2` at session creation. Deliberately **not** `readonly=1`: level 1 rejects `file()`/`s3()`/`url()` (chDB's core use case), level 2 allows reads + those functions while blocking `INSERT`/`CREATE`/`ALTER`/`DROP`. Can't be escaped (`SET readonly=0` and multi-statement writes both fail). Opt out with `read_only=False`.
2. **Value binding** — values pass as chDB params (`{name:Type}` + dict), never concatenated; identifiers go through a shared `quote_ident`.
3. **Result caps** — `max_rows`/`max_bytes`; truncation is flagged (`truncated`), never silent. 64-bit ints kept exact.
4. **Error-to-model** — `call()` returns `{ok, error:{code,type,message}}` for agents; methods raise typed `ChDBError`. Classification (`Code: N … (TYPE)`) is shared across languages.

## Cross-language consistency mechanism

`conformance/cases.jsonl` is a language-neutral list of setup-free cases (they use `numbers()`, a shipped CSV fixture, `system.*`, params, and write-blocking DDL) with expected outcomes. Each binding ships a thin (~30 LOC) runner that loads the same file and asserts. A binding is "contract-conformant" when every case passes; intentional divergences are added as documented cases, never silent. `conformance/README.md` explains the format and how `chdb-node` consumes it.

## Files

- `chdb/agents/{__init__,tool,errors,safety}.py` — implementation
- `chdb/agents/CONTRACT.md`, `chdb/agents/conformance/{cases.jsonl,README.md,fixtures/sample.csv}` — the shared contract + fixture
- `tests/test_agents_conformance.py` — reference runner over the fixture
- `tests/test_agents_chdbtool.py` — Python-specific unit tests
- `chdb/__init__.py` — register `agents`

## Tests

Verified against the `chdb-core` engine (both the published `chdb 4.1.8` / CH 26.3.9.1 and a local `chdb-core` main build / CH 26.5.1.1):
- `python -m unittest tests.test_agents_conformance` — 13 conformance cases pass (incl. the `file()`-under-read-only canary that would fail on `readonly=1`).
- `python -m unittest tests.test_agents_chdbtool` — 11 unit tests pass (error parser, `quote_ident`, `tool_specs`, read-only opt-out + writable, `aquery`).

## Follow-ups (not in this PR)

- `chdb-node`: add a runner over `conformance/cases.jsonl` (vendored/submoduled) and align `chdbTools()` to the contract.
- Ship framework shims (CrewAI / Pydantic AI / AutoGen / smolagents / LangChain) as thin wrappers over `ChDBTool`.
- Optionally extract the shared safety/error primitives so mcp-clickhouse and chdb-node import rather than re-derive them.

> [!NOTE]
> ### Add `chdb.agents.ChDBTool` as a canonical agent tool for querying chDB from LLM agents
> - Adds a new `chdb.agents` subpackage exposing `ChDBTool`, a session-backed query tool designed for use in LLM agent frameworks, with sync and async query APIs, schema introspection, DataFrame querying, and tool-dispatch via `call()`.
> - Enforces read-only mode (`readonly=2`) by default; supports optional file allowlists, execution timeouts, row/byte caps with truncation flags, and pre-attached file views via `CREATE VIEW`.
> - Adds structured error types (`ChDBError`, `ChDBReadOnlyError`, `ChDBSyntaxError`, `ChDBUnknownObjectError`) parsed from ClickHouse error messages, with `to_dict()` for agent-consumable error envelopes.
> - Adds safety helpers in [`safety.py`](https://github.com/chdb-io/chdb/pull/594/files#diff-a22a8a09033bc521b871b01b4f120aa121bac4beed603e3748b4801c858f3537) for identifier/string quoting and file-path scanning against an allowlist.
> - Includes a cross-language contract ([`CONTRACT.md`](https://github.com/chdb-io/chdb/pull/594/files#diff-911bffae59a61a359083551d3c4cd6855e8839e1f73d500396503aac7c9b8298)), conformance JSONL fixtures, and unit tests covering error classification, read-only enforcement, async parity, and DataFrame queries.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 89ed79c.</sup>
> 
>


ShawnChen-Sirius:
@chibugai, review it

chdb/agents/errors.py
```diff
@@ -0,0 +1,87 @@
...
+    "ChDBError",
+    "ChDBReadOnlyError",
+    "ChDBSyntaxError",
+    "ChDBUnknownObjectError",
+    "parse_error",
+]
+
+# `Code: N. DB::Exception: <msg>. (TYPE)` — TYPE is the trailing (UPPER_SNAKE).
+_ERR_RE = re.compile(
+    r"Code:\s*(?P<code>\d+)\.\s*DB::Exception:\s*(?P<msg>.*?)\.?\s*\((?P<type>[A-Z0-9_]+)\)",
```
chibugai: The non-greedy `(?P<msg>.*?)` will stop at the **first** `(UPPER_SNAKE)` token it encounters, so any ClickHouse message whose body contains a parenthesized upper-snake token before the real trailing `(TYPE)` will mis-parse. For example `Code: 62. DB::Exception: Cannot parse expression near (SOME_TOKEN) here. (SYNTAX_ERROR)` would capture `type=SOME_TOKEN` and truncate `msg`. The trailing type is the authoritative one; consider anchoring it to end-of-string, e.g. append `\s*$` and/or use a greedy `(?P<msg>.*)` so the last parenthesized token is matched.

```suggestion
    r"Code:\s*(?P<code>\d+)\.\s*DB::Exception:\s*(?P<msg>.*?)\.?\s*\((?P<type>[A-Z0-9_]+)\)\s*$",
```

chdb/agents/tool.py
```diff
@@ -0,0 +1,240 @@
...
+        never formatted into `sql`. Returns a `QueryResult`; raises `ChDBError`.
+        """
+        if not isinstance(sql, str) or sql.strip() == "":
+            raise ChDBError("sql must be a non-empty string")
+        cap = self.max_rows if max_rows is None else max(1, int(max_rows))
+        try:
+            res = self._session.query(sql, "JSON", params=params or {})
+        except Exception as e:  # engine failure -> typed error
+            raise parse_error(e)
+        obj = json.loads(res.bytes().decode() or "{}")
```
chibugai: The `json.loads(...)` call is outside the `try/except` above, so if the engine returns non-JSON or malformed output (e.g. an edge-case statement or empty body that isn't `{}`), a raw `json.JSONDecodeError` propagates instead of the typed `ChDBError` that this method's docstring promises. Consider moving the decode/parse inside the try (or wrapping it) so parse failures are surfaced as `ChDBError`.

chdb/agents/tool.py
```diff
@@ -0,0 +1,240 @@
...
+        data = obj.get("data", []) or []
+        meta = obj.get("meta", []) or []
+        stats = obj.get("statistics", {}) or {}
+        cols = [m.get("name") for m in meta]
+        truncated = len(data) > cap
+        rows = data[:cap] if truncated else data
+        # secondary byte guard: trim further if the serialized slice is large
+        if not truncated and self.max_bytes:
```
chibugai: The byte guard only runs when the row cap was NOT already hit (`if not truncated`). If a query returns fewer rows than `max_rows` but each row is large, or if the row cap is high, a result truncated by rows is never re-checked against `max_bytes`, so the returned payload can still exceed `max_bytes`. Consider always applying the byte guard so `max_bytes` is enforced regardless of the row-cap outcome.

chdb/agents/tool.py
```diff
@@ -0,0 +1,240 @@
...
+            size = 0
+            for i, r in enumerate(rows):
+                size += len(json.dumps(r, separators=(",", ":")))
+                if size > self.max_bytes:
+                    rows = rows[:i]
+                    truncated = True
+                    break
+        return QueryResult(rows, truncated, cols, stats.get("elapsed"), stats.get("rows_read"))
```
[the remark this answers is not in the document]
ShawnChen-Sirius: Fixed in ce15ab2 — `bytes_read` now reads `statistics.bytes_read` (was `rows_read`).

chdb/agents/tool.py
```diff
@@ -0,0 +1,240 @@
...
+        self.elapsed_s = elapsed_s
+        self.bytes_read = bytes_read
+
+    def to_dict(self):
+        return {
+            "rows": self.rows,
+            "row_count": self.row_count,
+            "truncated": self.truncated,
+            "column_names": self.column_names,
+        }
```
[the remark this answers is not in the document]
ShawnChen-Sirius: Fixed in ce15ab2 — `QueryResult.to_dict()` now includes `elapsed_s` and `bytes_read`.

chdb/agents/errors.py
ShawnChen-Sirius: Fixed in ce15ab2 — the type regex now matches the last `(UPPER_SNAKE)` (greedy msg), so a parenthesized token in the message body can't hijack the type. Added a regression test.

chdb/agents/tool.py
ShawnChen-Sirius: Fixed in ce15ab2 — the JSON decode is now inside the `try`, so malformed/non-JSON engine output becomes a typed `ChDBError` instead of a bare `JSONDecodeError`.

chdb/agents/tool.py
ShawnChen-Sirius: Fixed in ce15ab2 — the `max_bytes` guard now runs even when the row cap already fired.

PR #596, opened 2026-07-01
Zero-copy Arrow interop (to_arrow / __arrow_c_stream__ / from_arrow)

### LINKED ISSUE #595: DataStore: zero-copy Arrow interop (to_arrow / __arrow_c_stream__ / from_arrow)
## Problem

To hand DataStore results to Polars / DuckDB / pandas 3.0 today, you must go through `to_df()`. That forces an Arrow → pandas → Arrow round-trip (chDB already emits Arrow internally) and degrades ClickHouse types via numpy (e.g. `UInt64` → `int64`, nullable ints → float). There is no Arrow export at all, and no standard way for Arrow consumers to pull from a DataStore.

## Proposal

Add Arrow interop to `DataStore`, built on the Arrow PyCapsule interface (the pandas 3.0 interchange standard):

- **`to_arrow(types='native'|'pandas')`** — return a `pyarrow.Table`. For a pipeline that fully pushes down to one chDB SQL query, ask chDB for `ArrowTable` output directly, so results keep ClickHouse-native Arrow types with no pandas round-trip. Falls back to converting the pandas result for pipelines that need pandas-side ops.
- **`__arrow_c_stream__()`** — expose the DataStore via the Arrow PyCapsule interface so `pa.table(ds)`, `pl.from_arrow(ds)`, `duckdb.sql('... FROM ds')` and `pd.DataFrame.from_arrow(ds)` work with no glue code.
- **`from_arrow(data)`** — build a DataStore from any PyCapsule producer (pyarrow Table/RecordBatch/reader, Polars, pandas 3.x, DuckDB, another DataStore).

Makes chDB ↔ pandas / Polars / DuckDB interchange free. PR to follow.

wudidapaopao:
Adds Arrow interop to `DataStore`, built on the Arrow PyCapsule interface (the pandas 3.0 standard). Lets DataStore results flow to Polars / DuckDB / pandas 3.0 without a pandas round-trip.

## What
- **`to_arrow(types='native'|'pandas')`** — a fully pushed-down pipeline returns a `pyarrow.Table` straight from chDB's `ArrowTable` output (no pandas round-trip), keeping ClickHouse-native types (`UInt64`→`uint64`, nullable ints stay nullable). Falls back to converting the pandas result for pipelines that need pandas-side ops.
- **`__arrow_c_stream__()`** — `pa.table(ds)`, `pl.from_arrow(ds)`, `duckdb.sql('... FROM ds')` and `pd.DataFrame.from_arrow(ds)` all work with no glue code.
- **`from_arrow(data)`** — ingest any PyCapsule producer (pyarrow Table/RecordBatch/reader, Polars, pandas 3.x, DuckDB, another DataStore).
- Adds `Connection.query_arrow` / `Executor.query_arrow`.

## Notes
- Additive only; `to_df()` and existing behavior unchanged.
- Arrow has no row index, so group keys from `groupby(as_index=True)` appear as regular columns.
- Ingest goes through pyarrow (like pandas 3.0's own `from_arrow`), so it is not zero-copy on import; the zero-copy/native-typed path is on export.

## Tests
`datastore/tests/test_arrow_interop.py` — 23 tests: native types, filter/select/groupby pushdown, interop (pyarrow/Polars/DuckDB/pandas), ingest, round-trips. Passing locally.

Closes #595

> [!NOTE]
> ### Add zero-copy Arrow interop (`to_arrow`, `from_arrow`, `__arrow_c_stream__`) to `DataStore`
> - Adds `DataStore.to_arrow(types='native'|'pandas')` which returns a `pyarrow.Table`; in `'native'` mode it pushes down a single SQL query to chDB's Arrow output format, falling back to pandas conversion when pushdown is unavailable.
> - Adds `DataStore.from_arrow(data)` classmethod accepting Arrow PyCapsule objects, `pyarrow.Table`/`RecordBatch`/`RecordBatchReader`, and pandas/Polars DataFrames.
> - Implements `DataStore.__arrow_c_stream__` so `DataStore` objects are directly consumable by PyArrow, Polars, DuckDB, and pandas 3.x without intermediate copies.
> - Adds `Connection.query_arrow` and `Executor.query_arrow` to execute SQL and return a `pyarrow.Table` via chDB's `ArrowTable` output format.
> - A new test module covers round-trips, native type preservation (decimal, list, map, struct, binary), empty/null columns, `requested_schema` casting, and error handling.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 54b97e9.</sup>
> 
>

### DIFF
--- a/datastore/connection.py
+++ b/datastore/connection.py
@@ -173,7 +173,12 @@ class Connection:
             raise ConnectionError("Not connected. Call connect() first.")
 
         self._log_query(sql, "Connection", "ArrowTable")
-        return self._conn.query(sql, "ArrowTable")
+        try:
+            return self._conn.query(sql, "ArrowTable")
+        except Exception as e:
+            self._logger.error("[chDB] Arrow query failed: %s", e)
+            friendly_msg = translate_remote_error(e)
+            raise ExecutionError(f"Query execution failed: {friendly_msg}\nSQL: {sql}")
 
     def query_df(
         self,
--- a/datastore/connection.py
+++ b/datastore/connection.py
@@ -155,6 +155,26 @@ class Connection:
             friendly_msg = translate_remote_error(e)
             raise ExecutionError(f"Query execution failed: {friendly_msg}\nSQL: {sql}")
 
+    def query_arrow(self, sql: str):
+        """
+        Execute a SQL query and return a ``pyarrow.Table`` directly from chDB.
+
+        Unlike :meth:`execute` (which materializes a pandas DataFrame via the
+        ``Dataframe`` output format), this requests chDB's ``ArrowTable`` output,
+        preserving ClickHouse's native Arrow types with no pandas round-trip.
+
+        Args:
+            sql: SQL query string
+
+        Returns:
+            pyarrow.Table
+        """
+        if self._conn is None:
+            raise ConnectionError("Not connected. Call connect() first.")
+
+        self._log_query(sql, "Connection", "ArrowTable")
+        return self._conn.query(sql, "ArrowTable")
+
     def query_df(
         self,
         sql: str,
--- a/datastore/core.py
+++ b/datastore/core.py
@@ -3618,7 +3618,15 @@ class DataStore(PandasCompatMixin):
             if table is not None:
                 return table
 
-        return pa.Table.from_pandas(self._execute(), preserve_index=False)
+        # Fallback: convert the pandas result. Arrow has no row index, so surface
+        # a meaningful index (a named index, or group keys from
+        # groupby(as_index=True), or a MultiIndex) as regular columns rather than
+        # dropping it — matching the native path and the docstring Note. A plain
+        # unnamed RangeIndex carries no data and is left out.
+        df = self._execute()
+        if df.index.names != [None]:
+            df = df.reset_index()
+        return pa.Table.from_pandas(df, preserve_index=False)
 
     def __arrow_c_stream__(self, requested_schema=None):
         """
--- a/datastore/core.py
+++ b/datastore/core.py
@@ -2802,7 +2802,8 @@ class DataStore(PandasCompatMixin):
         (``__arrow_c_stream__`` / ``__arrow_c_array__``) — a ``pyarrow.Table``,
         ``polars.DataFrame``, a pandas 3.x DataFrame, another DataStore, a DuckDB
         relation, etc. — as well as raw pyarrow ``Table`` / ``RecordBatch`` /
-        ``RecordBatchReader`` objects.
+        ``RecordBatchReader`` objects, and a plain ``pandas.DataFrame`` (which on
+        pandas 2.x has no PyCapsule interface, so it is handled directly).
 
         Like pandas 3.0's ``DataFrame.from_arrow``, ingest currently relies on
         pyarrow to materialize a pandas frame, so it is *not* zero-copy on import.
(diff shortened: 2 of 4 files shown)


wudidapaopao:
@chibugai, review this PR.

datastore/connection.py
```diff
@@ -155,6 +155,26 @@ def execute(self, sql: str, output_format: str = "Dataframe") -> "QueryResult":
...
+            sql: SQL query string
+
+        Returns:
+            pyarrow.Table
+        """
+        if self._conn is None:
+            raise ConnectionError("Not connected. Call connect() first.")
+
+        self._log_query(sql, "Connection", "ArrowTable")
+        return self._conn.query(sql, "ArrowTable")
```
chibugai: Unlike `execute` and `query_df`, this method does not wrap the `_conn.query()` call in a try/except. On failure, raw chDB exceptions propagate untranslated instead of being converted into a friendly `ExecutionError`. This breaks the established error-handling pattern and exposes low-level errors to callers. Consider wrapping the call consistently:

```python
try:
    return self._conn.query(sql, "ArrowTable")
except Exception as e:
    self._logger.error("[chDB] Arrow query failed: %s", e)
    friendly_msg = translate_remote_error(e)
    raise ExecutionError(f"Query execution failed: {friendly_msg}\nSQL: {sql}")
```

datastore/core.py
```diff
@@ -3501,6 +3569,164 @@ def to_pandas(self) -> pd.DataFrame:
...
+            raise ValueError(
+                f"to_arrow(types=...) must be 'native' or 'pandas', got {types!r}"
+            )
+
+        if types == "native":
+            table = self._try_native_arrow()
+            if table is not None:
+                return table
+
+        return pa.Table.from_pandas(self._execute(), preserve_index=False)
```
chibugai: The fallback uses `preserve_index=False`, which **drops** the index instead of converting it to columns. This contradicts the docstring Note ("group keys from `groupby(as_index=True)` appears as regular columns here") and silently loses data: a `groupby(as_index=True).agg(...)` pipeline routed through this fallback (e.g. a DataFrame source, or any pandas-side op, or `types="pandas"`) returns a Table missing the group-key columns. This also diverges from the native path (`_try_native_arrow`), where group keys stay as columns. Consider resetting the index into columns before converting, e.g.:

    df = self._execute()
    if df.index.name is not None or isinstance(df.index, pd.MultiIndex) or df.index.names != [None]:
        df = df.reset_index()
    return pa.Table.from_pandas(df, preserve_index=False)

(or use `preserve_index=True` only for named/group indexes) so the fallback matches both the documented behavior and the native path.

datastore/connection.py
wudidapaopao: Good catch — fixed in 54b97e9. `query_arrow` now wraps the `_conn.query()` call in the same try/except as `execute()`, logging and translating failures via `translate_remote_error` into a friendly `ExecutionError` instead of leaking the raw chDB exception. Added a regression test (`TestQueryArrowErrorHandling`) that asserts an invalid query raises `ExecutionError`.

datastore/core.py
wudidapaopao: Agreed, this was a real data-loss bug — fixed in 54b97e9. The fallback now resets a *meaningful* index into columns before conversion:

```python
df = self._execute()
if df.index.names != [None]:
    df = df.reset_index()
return pa.Table.from_pandas(df, preserve_index=False)
```

So group keys from `groupby(as_index=True)` (and any named index / MultiIndex) are preserved as columns, matching both the native path and the docstring Note. A plain unnamed `RangeIndex` is still left out (no spurious index column). Confirmed the repro (`from_df(...).groupby("city").agg(...)` previously returned only `["score"]`, now `["city", "score"]`). Added regressions for the from_df-groupby, `types="pandas"`-groupby, and named-index cases.

wudidapaopao:

ShawnChen-Sirius (approved):
LGTM, approve to merge.

ShawnChen-Sirius:

PR #604, opened 2026-07-13
ci: pin pyarrow<25 in Linux arm64 DataStore step

### LINKED ISSUE #603: Linux arm64 CI: DataStore test_tail_uses_sql_offset_limit fails on main — pyarrow 25.0.0 aarch64 parquet read corruption
### Failing runs

- https://github.com/chdb-io/chdb/actions/runs/29224196603 — `Build & Test Linux arm64` / `Test DataStore (Linux arm64)`, push to `main` @ 83c8d00b (merge of #602), 2026-07-13T04:32Z — **failed on both attempt 1 and attempt 2** (re-run by the PR author at 04:54Z).
- Failing step: *Test DataStore on all Python versions with pandas 2.x and 3.x*, at the **Python 3.10 + pandas 2.3.3 + pyarrow 25.0.0** combo. The Python 3.9 combo (pyarrow 21.0.0 — pyarrow 25 ships no cp39 wheels) passes the full suite in the same job both times.

### Signature

`datastore/tests/test_head_performance.py::TestTailPerformanceOptimization::test_tail_uses_sql_offset_limit` — parquet round-trip corruption, Linux arm64 only.

**Attempt 1** — `pd.read_parquet()` of the file the test just wrote with `big_df.to_parquet()` raises:

```
tests/test_head_performance.py:329: in test_tail_uses_sql_offset_limit
    pd_df = pd.read_parquet(parquet_path)
pandas/io/parquet.py:669: in read_parquet
pandas/io/parquet.py:265: in read
pyarrow/parquet/core.py:1964: in read_table
E   pyarrow.lib.ArrowInvalid: Index not in dictionary bounds
```

**Attempt 2** — the read "succeeds" but returns **silently corrupted data**; the comparison against chdb then fails:

```
E   Column 'a' values don't match
E    ACTUAL: array([95, 96, 97, 98, 99])   <- chdb (DataStore tail via OFFSET/LIMIT) — CORRECT
E    DESIRED: array([ 0, 96, 97, 98, 99])  <- pandas read_parquet().tail(5) — row 95 read as 0
```

Note the file (`'a': range(100)`, `'b': 100 distinct strings`) is written **and** read by pandas/pyarrow; chdb only reads it. chdb's native parquet reader returns the correct values in attempt 2 — if the file itself were corrupt at row 95, chdb's `OFFSET 95 LIMIT 5` would return the same wrong value. So this is a **pyarrow read-side decode bug**, with two manifestations of one defect: a misread dictionary index that is either out-of-bounds (attempt 1, `ArrowInvalid`) or in-bounds-but-wrong (attempt 2, `dict[0]` = 0 instead of `dict[95]` = 95).

### Root cause

**pyarrow 25.0.0 (`manylinux_2_28_aarch64` wheel) nondeterministically mis-decodes dictionary-encoded parquet columns.**

Version differential evidence:

- Last green `main` run (2026-07-10T05:26Z, run 29071356055): py3.10 combo installed **pyarrow 24.0.0** — full suite green.
- pyarrow **25.0.0 was published 2026-07-10T08:25Z**, ~3 h later. This run is the **first Linux arm64 run since** (the DataStore step installs unpinned latest pyarrow as a dependency of the chdb wheel).
- **macOS arm64** (also aarch64, `macosx` wheel) on the same commit 83c8d00b: full DataStore suite **passes** with pyarrow 25.0.0 on py3.10/3.11/3.12/3.13 × pandas 2.x/3.x → not generic-aarch64, specific to the manylinux aarch64 wheel (or its interaction with the self-hosted runner's CPU).
- **Local x86_64** (chdb 3.7.0 + pandas 2.3.3 + pyarrow 25.0.0, py3.12): 300 iterations of the exact write→read round-trip pure, plus 300 iterations with chdb loaded in-process and querying the same file — **0 failures**.
- Merged PR #602 touches `chdb/agents/*` + agent tests only — unrelated to this code path.

Local repro on Linux arm64 was not possible (no arm64 host available to me); the two CI attempts serve as the reproduction (2/2).

### Impact

- Every `Build & Test Linux arm64` run on main/PRs will roll these dice at this test (100-entry dictionary appears to be the trigger; smaller frames in other tests don't hit it) — expect persistent red.
- The attempt-2 manifestation is **silent wrong data** from `pd.read_parquet` on Linux aarch64 + pyarrow 25.0.0 — worth flagging beyond CI: any user on that platform can hit it, independent of chdb.

### Suggested fix

1. Pin `"pyarrow<25"` in the DataStore test-step pip install of `build_linux_arm64_wheels-gh.yml` until an upstream fix (25.0.1) lands — PR follows.
2. Consider reporting upstream to apache/arrow (I found no existing report for `Index not in dictionary bounds` on 25.0.0 aarch64).

### Dedup

Searched open+closed issues for `Index not in dictionary bounds`, `test_tail_uses_sql_offset_limit`, `test_head_performance`, `ArrowInvalid` — no match (#94/#120 are old, unrelated ArrowInvalid errors).

chibugai:
- CI Fix or Improvement (changelog entry is not required)

Pin pyarrow<25 in the Linux arm64 DataStore CI step to avoid a pyarrow 25.0.0 aarch64 parquet decode bug.

### Fixes

Fixes #603

### Reproduction

`Build & Test Linux arm64` on `main` @ 83c8d00b failed **2/2 attempts** at
`datastore/tests/test_head_performance.py::TestTailPerformanceOptimization::test_tail_uses_sql_offset_limit`
on the Python 3.10 + pandas 2.3.3 + **pyarrow 25.0.0** combo:
run https://github.com/chdb-io/chdb/actions/runs/29224196603 (attempts 1 and 2).

- Attempt 1: `pyarrow.lib.ArrowInvalid: Index not in dictionary bounds` from `pd.read_parquet()` on a file pandas itself just wrote.
- Attempt 2: silent corruption — `pd.read_parquet().tail(5)` returns `[0, 96, 97, 98, 99]` while chdb reading the same file returns the correct `[95, 96, 97, 98, 99]`.

### Root cause

pyarrow 25.0.0's `manylinux_2_28_aarch64` wheel (published 2026-07-10T08:25Z, three hours after the last green arm64 run, which used 24.0.0) nondeterministically mis-decodes dictionary-encoded parquet columns: the misread index is either out of bounds (attempt 1) or valid-but-wrong (attempt 2: `dict[0]` instead of `dict[95]`). chdb's native parquet reader is not affected — if the file were corrupt on disk, chdb's `OFFSET 95 LIMIT 5` would return the same wrong value; it does not. Full analysis in #603.

Not reproducible on macOS arm64 (same commit, same pyarrow 25.0.0, all combos green) nor on Linux x86_64 (600 local iterations of the exact round-trip, with and without chdb loaded in-process, zero failures) — the pin is therefore scoped to the Linux arm64 workflow only, so the other platforms keep exercising latest pyarrow.

### Fix

Add `"pyarrow<25"` to the DataStore test-step pip install in `build_linux_arm64_wheels-gh.yml` (this line runs after the wheel install, so it also downgrades the transitively-resolved pyarrow). Comment links #603 so the pin can be dropped when a fixed upstream release (e.g. 25.0.1) is out. No effect on the py3.9 combo, which already resolves to pyarrow 21.0.0 (no cp39 wheels in 25.x).

### Regression test

The `Test DataStore (Linux arm64)` job itself: with the pin it resolves pyarrow 24.0.0 on py>=3.10 — the exact configuration of the last green main run (29071356055). YAML validated locally; the change cannot be exercised outside CI (no arm64 host available to me).

> [!NOTE]
> ### Pin `pyarrow<25` in Linux arm64 DataStore CI test step
> Adds a `pyarrow<25` version constraint to the pip install command in the [Linux arm64 wheels workflow](https://github.com/chdb-io/chdb/pull/604/files#diff-fd24c962953455ae96b49526dd5a9f87b30c58a054fa3be36afa304ba1f55359) to work around an upstream incompatibility. Three comments are added inline to explain the pin and reference the upstream issue.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 48ec270.</sup>
> 
>

### DIFF
--- a/.github/workflows/build_linux_arm64_wheels-gh.yml
+++ b/.github/workflows/build_linux_arm64_wheels-gh.yml
@@ -117,7 +117,10 @@ jobs:
 
               pyenv shell $version
               python -m pip install dist/*.whl --force-reinstall
-              python -m pip install pytest pytest-cov hypothesis ray "$PANDAS_CONSTRAINT"
+              # pyarrow 25.0.0 manylinux aarch64 wheel mis-decodes dictionary-encoded
+              # parquet columns (crash or silent wrong data); pin until fixed upstream.
+              # See https://github.com/chdb-io/chdb/issues/603
+              python -m pip install pytest pytest-cov hypothesis ray "$PANDAS_CONSTRAINT" "pyarrow<25"
               echo "Installed pandas version:"
               python -c "import pandas; print(pandas.__version__)"
 


chibugai:
Superseded by #605, which pins pyarrow<25 in both the pyenv setup step and the DataStore test step (this PR only covered the test step). Closing in favor of the merged fix. Will confirm on #603 once the arm64 run on ce71a518 goes green.

PR #611, opened 2026-07-14
Enable external sort by default and restore the server memory limit

### LINKED ISSUE #610: Large ORDER BY fails with MEMORY_LIMIT_EXCEEDED instead of spilling to disk
DataStore's default chdb settings set `max_server_memory_usage_to_ram_ratio=0`, disabling the server memory limit. Without it, OOM ends in a kernel kill of the whole Python process instead of a catchable error, and the engine's ratio-based auto-spill for GROUP BY / JOIN never engages (it derives its thresholds from that limit).

Independently, ClickHouse ships `max_bytes_before_external_sort=0` and the ratio setting alone can never trigger an external sort, so a large ORDER BY always fails with MEMORY_LIMIT_EXCEEDED even when spilling could complete it. Verified in a 1GB cgroup: a 1.6GB sort fails at the limit with zero `ExternalSort*` events.

Proposal:
- Restore `max_server_memory_usage_to_ram_ratio=0.9` (engine default).
- At connect time, set `max_bytes_before_external_sort` to half of the effective server memory limit — the officially recommended value and the ClickHouse Cloud default.

Validated: the same 1.6GB sort in a 1GB cgroup completes in 65s (4 spill parts); on a 123GB box a 4GB sort shows no spill and no perf change (329s vs 331s baseline) since the ratio valve keeps sorts in memory until real pressure.

wudidapaopao:
Closes #610

- Restore `max_server_memory_usage_to_ram_ratio=0.9` in the default chdb settings, re-enabling the catchable server-level memory limit and the ratio-based auto-spill for GROUP BY / JOIN.
- After connect, set `max_bytes_before_external_sort` to half of the effective server memory limit (official recommendation, ClickHouse Cloud default) so large ORDER BY spills to disk instead of failing with MEMORY_LIMIT_EXCEEDED. A user-provided value via `set_chdb_setting` or connection params is applied as-is; if the server limit is disabled, nothing is set.

Validation: in a 1GB cgroup a 1.6GB sort previously failed with MEMORY_LIMIT_EXCEEDED and now completes in 65s (4 spill parts); on a 123GB box a 4GB sort runs unchanged (329s vs 331s baseline, no spill) because the default ratio valve keeps sorts in memory until real memory pressure.

Unit tests cover the new defaults, the derived sort threshold, user overrides, and the disabled-limit fallback.

> [!NOTE]
> ### Enable external sort by default and restore server memory limit to 0.9 of RAM
> - Changes the default `max_server_memory_usage_to_ram_ratio` from `0` to `0.9` in [datastore/config.py](https://github.com/chdb-io/chdb/pull/611/files#diff-5cac076f9632371fd2a4e9a35a94846c51fa8b681451866351fb966c446e4a64), restoring a meaningful server memory limit.
> - Adds `Connection._apply_external_sort_default()` in [datastore/connection.py](https://github.com/chdb-io/chdb/pull/611/files#diff-5bedd723d353a154f038b0c67bbe3900ae887c8c88fa31837ba3bb34a585ee33) that sets `max_bytes_before_external_sort` to half the server memory limit after each connection is established.
> - User-supplied overrides for `max_bytes_before_external_sort` (via global settings or per-instance params) are applied as-is; boolean overrides are rejected with a `ValueError`.
> - Risk: if applying the external sort setting fails (e.g. invalid override value), `connect()` now aborts and leaves `_conn` as `None` instead of succeeding silently.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized c7f2e48.</sup>
> 
>

### DIFF
--- a/datastore/connection.py
+++ b/datastore/connection.py
@@ -115,10 +115,13 @@ class Connection:
         settings = {**get_chdb_settings(), **self.connection_params}
         user_value = settings.get('max_bytes_before_external_sort')
         if user_value is not None:
+            if isinstance(user_value, bool):
+                raise ValueError(f"Invalid max_bytes_before_external_sort value: {user_value!r}")
             if isinstance(user_value, (int, float)):
                 self._conn.query(f"SET max_bytes_before_external_sort = {int(user_value)}")
             else:
-                self._conn.query(f"SET max_bytes_before_external_sort = '{user_value}'")
+                escaped = str(user_value).replace("\\", "\\\\").replace("'", "\\'")
+                self._conn.query(f"SET max_bytes_before_external_sort = '{escaped}'")
             return
         try:
             raw = self._conn.query(
--- a/datastore/connection.py
+++ b/datastore/connection.py
@@ -96,10 +96,33 @@ class Connection:
             self._conn = chdb.connect(conn_str)
             if self._conn is not None:
                 _freeze_chdb_settings()
+                self._apply_external_sort_default()
             self._logger.debug("[chDB] Connected: %s", conn_str)
             return self
         except Exception as e:
             raise ConnectionError(f"Failed to connect to chdb: {e}")
+
+    def _apply_external_sort_default(self) -> None:
+        """Enable external sort with half of the server memory limit as threshold."""
+        settings = {**get_chdb_settings(), **self.connection_params}
+        user_value = settings.get('max_bytes_before_external_sort')
+        if user_value is not None:
+            if isinstance(user_value, (int, float)):
+                self._conn.query(f"SET max_bytes_before_external_sort = {int(user_value)}")
+            else:
+                self._conn.query(f"SET max_bytes_before_external_sort = '{user_value}'")
+            return
+        try:
+            raw = self._conn.query(
+                "SELECT value FROM system.server_settings WHERE name = 'max_server_memory_usage'",
+                'TSV',
+            )
+            server_limit = int(str(raw).strip() or 0)
+            if server_limit <= 0:
+                return
+            self._conn.query(f"SET max_bytes_before_external_sort = {server_limit // 2}")
+        except Exception as e:
+            self._logger.debug("[chDB] Could not apply external sort default: %s", e)
 
     def cursor(self):
         """Get a cursor for executing queries."""
--- a/datastore/connection.py
+++ b/datastore/connection.py
@@ -96,7 +96,15 @@ class Connection:
             self._conn = chdb.connect(conn_str)
             if self._conn is not None:
                 _freeze_chdb_settings()
-                self._apply_external_sort_default()
+                try:
+                    self._apply_external_sort_default()
+                except Exception:
+                    try:
+                        self._conn.close()
+                    except Exception:
+                        pass
+                    self._conn = None
+                    raise
             self._logger.debug("[chDB] Connected: %s", conn_str)
             return self
         except Exception as e:
--- a/datastore/tests/test_chdb_settings.py
+++ b/datastore/tests/test_chdb_settings.py
@@ -379,8 +379,9 @@ class TestExternalSortDefault:
         conn = Connection(':memory:')
         conn.connect()
         try:
-            value, _ = self._query_setting(conn, 'max_bytes_ratio_before_external_sort')
-            assert float(value) == 0.5
+            value, changed = self._query_setting(conn, 'max_bytes_ratio_before_external_sort')
+            assert changed == '0'
+            assert float(value) > 0
         finally:
             conn.close()
 
@@ -409,12 +410,9 @@ class TestExternalSortDefault:
 
         set_chdb_setting('max_bytes_before_external_sort', 'not_a_valid_size')
         conn = Connection(':memory:')
-        try:
-            with pytest.raises(DataStoreConnectionError, match="Failed to connect"):
-                conn.connect()
-        finally:
-            if conn._conn is not None:
-                conn.close()
+        with pytest.raises(DataStoreConnectionError, match="Failed to connect"):
+            conn.connect()
+        assert conn._conn is None
 
     def test_no_server_limit_skips_external_sort_default(self):
         _run_chdb_in_subprocess("""
@@ -427,11 +425,11 @@ class TestExternalSortDefault:
             conn.connect()
             try:
                 r = conn._conn.query(
-                    "SELECT value FROM system.settings "
+                    "SELECT changed FROM system.settings "
                     "WHERE name = 'max_bytes_before_external_sort'",
                     'TSV',
                 )
-                assert str(r).strip() == '0', f"Expected '0' but got: {r}"
+                assert str(r).strip() == '0', f"Expected setting to be unchanged but got: {r}"
             finally:
                 conn.close()
         """)
--- a/datastore/tests/test_chdb_settings.py
+++ b/datastore/tests/test_chdb_settings.py
@@ -32,7 +32,7 @@ def _reset_chdb_settings_state():
     _config_module._chdb_settings = {
         'memory_worker_correct_memory_tracker': 1,
         'max_server_memory_usage': 0,
-        'max_server_memory_usage_to_ram_ratio': 0,
+        'max_server_memory_usage_to_ram_ratio': 0.9,
     }
     _config_module._chdb_settings_frozen = False
 
@@ -74,7 +74,7 @@ class TestChdbSettingsDefaults:
         assert settings == {
             'memory_worker_correct_memory_tracker': 1,
             'max_server_memory_usage': 0,
-            'max_server_memory_usage_to_ram_ratio': 0,
+            'max_server_memory_usage_to_ram_ratio': 0.9,
         }
 
     def test_default_applied_to_connection(self):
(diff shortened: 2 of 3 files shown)


wudidapaopao:
@chibugai, review this PR.

datastore/connection.py
```diff
@@ -96,11 +96,34 @@ def connect(self) -> "Connection":
...
 
+    def _apply_external_sort_default(self) -> None:
+        """Enable external sort with half of the server memory limit as threshold."""
+        settings = {**get_chdb_settings(), **self.connection_params}
+        user_value = settings.get('max_bytes_before_external_sort')
+        if user_value is not None:
+            if isinstance(user_value, (int, float)):
+                self._conn.query(f"SET max_bytes_before_external_sort = {int(user_value)}")
+            else:
+                self._conn.query(f"SET max_bytes_before_external_sort = '{user_value}'")
```
chibugai: Unescaped string interpolation into SQL: if `user_value` is a string containing a single quote (or backslash), the generated `SET` statement will break or allow SQL injection into the session. Since this value can come from user-facing config (`set_chdb_setting` / connection params), escape it or validate its format (e.g., allow only digits with an optional size suffix like `1G`) before interpolation.

```suggestion
            else:
                escaped = str(user_value).replace("\\", "\\\\").replace("'", "\\'")
                self._conn.query(f"SET max_bytes_before_external_sort = '{escaped}'")
```

datastore/connection.py
```diff
@@ -96,11 +96,34 @@ def connect(self) -> "Connection":
             self._conn = chdb.connect(conn_str)
             if self._conn is not None:
                 _freeze_chdb_settings()
+                self._apply_external_sort_default()
```
chibugai: Resource leak: `_apply_external_sort_default()` can raise after `chdb.connect()` has already succeeded (e.g., an invalid user override, as exercised by `test_invalid_user_override_raises`). In that case `connect()` raises `ConnectionError` but `self._conn` is left open and never closed, forcing callers to manually clean up a "failed" connection. Close the underlying connection in the failure path before re-raising.

```suggestion
                try:
                    self._apply_external_sort_default()
                except Exception:
                    self._conn.close()
                    self._conn = None
                    raise
```

datastore/tests/test_chdb_settings.py
```diff
@@ -342,6 +346,107 @@ def test_override_setting_via_datastore_file_source(self, tmp_path):
...
+            conn.close()
+
+    def test_ratio_valve_kept_at_engine_default(self):
+        conn = Connection(':memory:')
+        conn.connect()
+        try:
+            value, _ = self._query_setting(conn, 'max_bytes_ratio_before_external_sort')
+            assert float(value) == 0.5
+        finally:
+            conn.close()
```
chibugai, line 389: This test pins the engine's built-in default for `max_bytes_ratio_before_external_sort` (0.5), which is not set anywhere by this codebase. This default has changed across ClickHouse versions (it was 0 before being changed to 0.5), so the test will break on a chdb engine upgrade even though the code under test is unchanged. Asserting `changed == '0'` (i.e., our code did not touch the ratio valve) is more robust and better expresses the intent 'kept at engine default'.

```suggestion
    def test_ratio_valve_kept_at_engine_default(self):
        conn = Connection(':memory:')
        conn.connect()
        try:
            _, changed = self._query_setting(conn, 'max_bytes_ratio_before_external_sort')
            assert changed == '0'
        finally:
            conn.close()
```

datastore/tests/test_chdb_settings.py
```diff
@@ -342,6 +346,107 @@ def test_override_setting_via_datastore_file_source(self, tmp_path):
...
+
+            conn = Connection(':memory:')
+            conn.connect()
+            try:
+                r = conn._conn.query(
+                    "SELECT value FROM system.settings "
+                    "WHERE name = 'max_bytes_before_external_sort'",
+                    'TSV',
+                )
+                assert str(r).strip() == '0', f"Expected '0' but got: {r}"
```
chibugai: Same engine-default brittleness: asserting the literal value '0' depends on the built-in default of `max_bytes_before_external_sort`, which can differ across chdb/ClickHouse versions. Since the intent is 'the derivation was skipped', querying `changed` and asserting it is '0' is version-independent and directly verifies the skip path.

```suggestion
                r = conn._conn.query(
                    "SELECT changed FROM system.settings "
                    "WHERE name = 'max_bytes_before_external_sort'",
                    'TSV',
                )
                assert str(r).strip() == '0', f"Expected setting to be unchanged but got: {r}"
```

datastore/tests/test_chdb_settings.py
```diff
@@ -342,6 +346,107 @@ def test_override_setting_via_datastore_file_source(self, tmp_path):
...
+            finally:
+                conn.close()
+        """)
+
+    def test_applied_via_global_executor(self):
+        executor = get_executor()
+        executor._ensure_connected()
```
chibugai, line 451: This test couples to the executor's private `_ensure_connected()` method. The executor exposes a public `execute()` API that establishes the connection internally (see `Executor.execute`), so using it keeps the test decoupled from internal implementation details and still exercises the real connection path.

```suggestion
    def test_applied_via_global_executor(self):
        executor = get_executor()
        executor.execute("SELECT 1")
```

datastore/tests/test_chdb_settings.py
```diff
@@ -342,6 +346,107 @@ def test_override_setting_via_datastore_file_source(self, tmp_path):
...
+        from datastore.exceptions import ConnectionError as DataStoreConnectionError
+
+        set_chdb_setting('max_bytes_before_external_sort', 'not_a_valid_size')
+        conn = Connection(':memory:')
+        try:
+            with pytest.raises(DataStoreConnectionError, match="Failed to connect"):
+                conn.connect()
+        finally:
+            if conn._conn is not None:
+                conn.close()
```
chibugai: Note: when `connect()` fails in `_apply_external_sort_default` (the invalid SET), `chdb.connect()` has already succeeded, `_freeze_chdb_settings()` has been called, and `conn._conn` still holds a live connection despite `connect()` raising. The test correctly compensates with the `if conn._conn is not None` cleanup, but this reveals the connection is left in a half-connected state on failure. Consider also asserting the post-failure state explicitly (or having the implementation reset `_conn` on failure) so this contract is documented rather than implicit.

datastore/connection.py
wudidapaopao: Not changing this one: the value comes from the same local, trusted config surface (`set_chdb_setting` / connection params) whose owner can already run arbitrary SQL on this embedded connection, so there is no privilege boundary to protect. A malformed value (quotes included) fails fast at connect time with a clear engine parse error, covered by `test_invalid_user_override_raises`.

datastore/connection.py
wudidapaopao: Good catch — fixed in ff3209f: the connection is closed and `_conn` reset to None before re-raising, and the test now asserts `conn._conn is None` after the failed connect.

datastore/tests/test_chdb_settings.py
wudidapaopao: Adopted with one tweak in ff3209f: assert `changed == '0'` as suggested, plus `float(value) > 0` — the non-zero ratio valve is a design assumption of the derived threshold (with the valve gone, an explicit byte threshold would make sorts spill unconditionally), so a future engine default of 0 should make this test fail loudly.

datastore/tests/test_chdb_settings.py
wudidapaopao: Fixed as suggested in ff3209f.

datastore/tests/test_chdb_settings.py
wudidapaopao: Keeping `_ensure_connected()` here: it is the established pattern in this file (e.g. `test_global_executor_uses_default_settings`), and the suggested replacement drops the assertions on the derived setting, which are the point of the test.

datastore/tests/test_chdb_settings.py
wudidapaopao: Addressed together with the resource-leak fix in ff3209f: `connect()` now closes and resets `_conn` on failure, and this test asserts `conn._conn is None` so the contract is explicit.

wudidapaopao:
Second-round comments addressed in c7f2e48:

- **connection.py:118 (bool)**: now rejected with a `ValueError` at connect time; covered by `test_bool_user_override_raises`.
- **connection.py:120 (escaping)**: single quotes and backslashes are now escaped before interpolation. Note this is for error clarity, not security — an embedded engine has no privilege boundary to cross (the config owner can already run arbitrary SQL on the same connection).
- **tests:352 / tests:104**: added explicit non-empty assertions so a missing setting fails with a clear message instead of an unpacking/parsing error.
- **tests:382**: dropped the `float(value) > 0` assertion; the test now only asserts `changed == '0'`, i.e. the invariant that datastore leaves the ratio valve untouched.
- **connection.py:115 (duplicated merge)**: keeping as-is — two call sites today; will extract a shared helper when a third appears.
- The flakiness note on tests:104 needs no change: server settings are baked at first connect (documented in the file docstring), and tests requiring non-default server settings already run in subprocesses.

wudidapaopao:

ShawnChen-Sirius (approved):
LGTM, approve to merge.

ShawnChen-Sirius:

PR #615, opened 2026-07-21
datastore: pandas 3.0 API parity — read_iceberg/to_iceberg, str.isascii, str.replace(dict), map(**kwargs)

### LINKED ISSUE #614: DataStore: missing pandas 3.0 APIs — read_iceberg/to_iceberg, str.isascii, str.replace(dict), Series.map(**kwargs)
A pandas 3.x compatibility investigation of DataStore found a handful of pandas 3.0 APIs that are still missing. The Arrow PyCapsule interop half (`to_arrow` / `__arrow_c_stream__` / `from_arrow`) is covered by #595 / #596; the remaining gaps:

1. **`read_iceberg` / `to_iceberg`** — pandas 3.0 added PyIceberg-backed IO; `datastore.read_iceberg` does not exist and `ds.to_iceberg` resolves to a lazy column reference instead of a method.
2. **`Series.str.isascii()`** — `AttributeError: 'StringAccessor' object has no attribute 'isascii'`.
3. **`Series.str.replace({pat: repl})`** — the pandas 3.0 dictionary form fails: `TypeError: _build_str_replace() missing 1 required positional argument: 'replacement'`.
4. **`Series.map(func, **kwargs)`** — pandas 3.0 forwards kwargs to the callable; `ColumnExpr.map() got an unexpected keyword argument`.

All four verified against pandas 3.0.3 (each works on a native pandas Series, fails on the DataStore mirror).

wudidapaopao:
Fixes #614

Fills the remaining pandas 3.0 API gaps from the compatibility investigation (the Arrow PyCapsule half is #596):

- **`read_iceberg()` / `DataStore.to_iceberg()`** — delegate to pandas' PyIceberg-backed IO and wrap the result in a DataStore; clear `NotImplementedError` on pandas < 3.
- **`Series.str.isascii()`** — registered as a STRING function (`match(s, '^[\x00-\x7F]*$')`), auto-injected into the accessor like the other `is*` predicates; empty strings are True, NULLs propagate.
- **`Series.str.replace({pat: repl})`** — dict form maps to chained `replaceAll`/`replaceRegexpAll`; `repl` alongside a dict raises `ValueError` like pandas; empty dict is identity; two-arg and regex forms unchanged.
- **`Series.map(func, **kwargs)`** — `ColumnExpr.map` and the executor fallback pass kwargs through to `Series.map` (pandas 3 forwards them to the callable).

## Tests

`datastore/tests/test_pandas3_api_gaps.py` — 16 tests: value + pandas-mirror checks for all four features, sequential-application semantics for dict replace, iceberg delegation contracts via monkeypatched pandas IO, and pandas-2 error guards.

- pandas 3.0.3: 14 passed, 2 skipped · pandas 2.3.3: 12 passed, 4 skipped
- regression: `test_pandas_compat` / `test_pandas_col_compat` / `test_from_dataframe` / `test_dataframe_sql_pushdown` (296 passed), string-accessor suites, exploratory string batches — all green; `ruff check datastore` clean

Note: `test_str_accessor_recursion_fix.py::test_str_extract_then_filter` hangs under pandas 3.0.3 identically on current `main` (A/B-checked in a clean worktree) — it is the known string+NULL engine issue (chdb-io/chdb-core#135 family), unrelated to this change.

> [!NOTE]
> ### Add pandas 3.0 API parity for `read_iceberg`, `to_iceberg`, `str.isascii`, `str.replace(dict)`, and `map(**kwargs)`
> - Adds `read_iceberg` in [pandas_api.py](https://github.com/chdb-io/chdb/pull/615/files#diff-09e41dc972d33e21863fddb5aabf4fb850999d2f20704e69224540c54f600659) and `to_iceberg` in [pandas_compat.py](https://github.com/chdb-io/chdb/pull/615/files#diff-9845c9fab6c8868000c586df009fb342cb663c720fa5f936e8a67880ca4f1780) to read/write Apache Iceberg tables via pandas 3.0; raises `NotImplementedError` on older pandas.
> - Adds `str.isascii()` support in [function_definitions.py](https://github.com/chdb-io/chdb/pull/615/files#diff-cbacf79f41ae1abc693cc41a19e6c190a39ee13c5ef773a4f49943dc56b2d066) using a regex match against `^[\x00-\x7F]*$` (empty strings return `True`, matching pandas/Python semantics).
> - Extends `str.replace` in [function_definitions.py](https://github.com/chdb-io/chdb/pull/615/files#diff-cbacf79f41ae1abc693cc41a19e6c190a39ee13c5ef773a4f49943dc56b2d066) to accept a `dict` for sequential pattern replacement, with validation that rejects mixed `replacement` args or non-string dict entries.
> - Extends `ColumnExpr.map` in [column_expr.py](https://github.com/chdb-io/chdb/pull/615/files#diff-cf897f1def888a8f77705c04e9b029521fd5aecb7f46dae2266113e0f3be2f6d) to forward `**kwargs` to the underlying pandas `Series.map`; raises `NotImplementedError` eagerly if kwargs are passed on pandas < 3.0.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 5e90170.</sup>
> 
>

### DIFF
--- a/datastore/column_expr.py
+++ b/datastore/column_expr.py
@@ -5685,6 +5685,21 @@ class ColumnExpr:
             Name: grade, dtype: float64
             >>> ds['x'].map(lambda v, add: v + add, add=10)  # pandas 3.0+
         """
+        if kwargs:
+            import inspect
+
+            # Fail at the call site on pandas < 3 (whose Series.map has no
+            # **kwargs), consistent with the eager NotImplementedError raised
+            # by the other pandas-3.0 additions, instead of an opaque
+            # TypeError at lazy execution time.
+            supports_kwargs = any(
+                p.kind is inspect.Parameter.VAR_KEYWORD
+                for p in inspect.signature(pd.Series.map).parameters.values()
+            )
+            if not supports_kwargs:
+                raise NotImplementedError(
+                    "Passing extra keyword arguments to map() requires pandas >= 3.0"
+                )
         return ColumnExpr(
             source=self,
             method_name="map",
--- a/datastore/column_expr.py
+++ b/datastore/column_expr.py
@@ -5663,13 +5663,16 @@ class ColumnExpr:
         # For scalar results or other types, can't use .get()
         return default
 
-    def map(self, arg, na_action=None) -> "ColumnExpr":
+    def map(self, arg, na_action=None, **kwargs) -> "ColumnExpr":
         """
         Map values of Series according to input mapping or function (lazy).
 
         Args:
             arg: Mapping correspondence (dict, Series, or function)
             na_action: If 'ignore', propagate NaN values without passing to mapping
+            **kwargs: Additional keyword arguments passed through to
+                pandas Series.map, which forwards them to ``arg`` when it is a
+                callable (pandas 3.0+)
 
         Returns:
             ColumnExpr: Lazy wrapper returning Series with mapped values
@@ -5680,12 +5683,13 @@ class ColumnExpr:
             1    3.0
             2    2.0
             Name: grade, dtype: float64
+            >>> ds['x'].map(lambda v, add: v + add, add=10)  # pandas 3.0+
         """
         return ColumnExpr(
             source=self,
             method_name="map",
             method_args=(arg,),
-            method_kwargs=dict(na_action=na_action),
+            method_kwargs=dict(na_action=na_action, **kwargs),
         )
 
     def fillna(
--- a/datastore/function_definitions.py
+++ b/datastore/function_definitions.py
@@ -275,16 +275,21 @@ def _build_str_replace(expr, pattern, replacement=None, regex: bool = False, ali
         # pandas 3.0: str.replace({'old': 'new', ...}); repl must not be given.
         if replacement is not None:
             raise ValueError("repl cannot be used when pat is a dictionary")
-        func_name = 'replaceRegexpAll' if regex else 'replaceAll'
         items = list(pattern.items())
+        for pat, repl in items:
+            # No silent str() coercion: pandas raises for e.g. {'a': None}, and
+            # stringifying would write literal 'None' into the data instead.
+            if not isinstance(pat, str) or not isinstance(repl, str):
+                raise TypeError("patterns and replacements in the dictionary must be strings")
         if not items:
             # Identity: single-argument coalesce returns the column unchanged.
             return Function('coalesce', expr, alias=alias)
+        func_name = 'replaceRegexpAll' if regex else 'replaceAll'
         result = expr
         for pat, repl in items[:-1]:
-            result = Function(func_name, result, Literal(str(pat)), Literal(str(repl)))
+            result = Function(func_name, result, Literal(pat), Literal(repl))
         last_pat, last_repl = items[-1]
-        return Function(func_name, result, Literal(str(last_pat)), Literal(str(last_repl)), alias=alias)
+        return Function(func_name, result, Literal(last_pat), Literal(last_repl), alias=alias)
 
     if replacement is None:
         raise TypeError("repl must be a string when pat is not a dictionary")
--- a/datastore/function_definitions.py
+++ b/datastore/function_definitions.py
@@ -248,25 +248,46 @@ def _build_replace(expr, to_replace, value=_REPLACE_SENTINEL, regex: bool = Fals
     func_type=FunctionType.SCALAR,
     category=FunctionCategory.STRING,
     aliases=['replaceAll', 'replace'],  # 'replace' alias for str accessor
-    doc='Replace substring occurrences. Maps to replace(s, from, to). Use regex=True for regex.',
+    doc='Replace substring occurrences. Maps to replace(s, from, to). Use regex=True for regex. '
+        'pat may be a dict of {pattern: replacement} (pandas 3.0), applied sequentially.',
 )
-def _build_str_replace(expr, pattern: str, replacement: str, regex: bool = False, alias=None):
+def _build_str_replace(expr, pattern, replacement=None, regex: bool = False, alias=None):
     """
     Replace substrings in a string column (pandas str.replace compatible).
 
     Args:
         expr: The expression to operate on
-        pattern: Substring pattern to find
-        replacement: Replacement string
+        pattern: Substring pattern to find, or a dict of {pattern: replacement}
+            (pandas 3.0 form; replacement must then be omitted)
+        replacement: Replacement string (required unless pattern is a dict)
         regex: If True, use regex replacement
         alias: Optional alias for the result
 
     Examples:
-        str.replace('a', 'b')     -> replaces all 'a' with 'b'
-        str.replace('a+', 'b', regex=True) -> regex replacement
+        str.replace('a', 'b')                -> replaces all 'a' with 'b'
+        str.replace('a+', 'b', regex=True)   -> regex replacement
+        str.replace({'a': 'b', 'c': 'd'})    -> sequential replacements
     """
     from .functions import Function
     from .expressions import Literal
+
+    if isinstance(pattern, dict):
+        # pandas 3.0: str.replace({'old': 'new', ...}); repl must not be given.
+        if replacement is not None:
+            raise ValueError("repl cannot be used when pat is a dictionary")
+        func_name = 'replaceRegexpAll' if regex else 'replaceAll'
+        items = list(pattern.items())
+        if not items:
+            # Identity: single-argument coalesce returns the column unchanged.
+            return Function('coalesce', expr, alias=alias)
+        result = expr
+        for pat, repl in items[:-1]:
+            result = Function(func_name, result, Literal(str(pat)), Literal(str(repl)))
+        last_pat, last_repl = items[-1]
+        return Function(func_name, result, Literal(str(last_pat)), Literal(str(last_repl)), alias=alias)
+
+    if replacement is None:
+        raise TypeError("repl must be a string when pat is not a dictionary")
 
     if regex:
         return Function('replaceRegexpAll', expr, Literal(pattern), Literal(replacement), alias=alias)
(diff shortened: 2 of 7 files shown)


wudidapaopao:
@chibugai, review this PR.

datastore/column_expr.py
```diff
@@ -5680,12 +5683,13 @@ def map(self, arg, na_action=None) -> "ColumnExpr":
...
             Name: grade, dtype: float64
+            >>> ds['x'].map(lambda v, add: v + add, add=10)  # pandas 3.0+
         """
         return ColumnExpr(
             source=self,
             method_name="map",
             method_args=(arg,),
-            method_kwargs=dict(na_action=na_action),
+            method_kwargs=dict(na_action=na_action, **kwargs),
         )
```
chibugai, line 5708: On pandas < 3.0, `Series.map` does not accept extra keyword arguments, so any `**kwargs` passed here will only fail later at lazy execution time with an opaque `TypeError: map() got an unexpected keyword argument ...`, far from the call site. Other pandas-3.0 additions in this change set (e.g., `read_iceberg`, `to_iceberg`) proactively detect the missing capability and raise a clear `NotImplementedError` at call time. For consistency and easier debugging, consider an eager version/capability check here when kwargs are supplied.
```suggestion
        if kwargs:
            import inspect

            if "kwargs" not in inspect.signature(pd.Series.map).parameters:
                raise NotImplementedError(
                    "Passing extra keyword arguments to map() requires pandas >= 3.0"
                )
        return ColumnExpr(
            source=self,
            method_name="map",
            method_args=(arg,),
            method_kwargs=dict(na_action=na_action, **kwargs),
        )
```

datastore/pandas_compat.py
```diff
@@ -2174,6 +2174,31 @@ def to_feather(self, path, **kwargs):
...
+            table_identifier: Iceberg table identifier, e.g. "namespace.table"
+            catalog_name: Name of the PyIceberg catalog to use
+            **kwargs: Additional DataFrame.to_iceberg() arguments
+                (catalog_properties, location, append, snapshot_properties)
+
+        Raises:
+            NotImplementedError: If the installed pandas has no
+                DataFrame.to_iceberg (requires pandas >= 3.0)
+        """
+        import pandas as pd
```
chibugai: Redundant local import: `pandas` is already imported as `pd` at the module level (line 26 of this file). The function-level `import pandas as pd` adds no value (there is no circular-import concern here) and is inconsistent with the other `to_*` methods in this mixin, which use the module-level import directly.
```suggestion
        if not hasattr(pd.DataFrame, "to_iceberg"):
```

datastore/function_executor.py
```diff
@@ -734,7 +734,10 @@ def _fillna_compat(s, value=None, method=None):
 
         # Apply and map
         self._pandas_implementations["apply"] = lambda s, func: s.apply(func)
-        self._pandas_implementations["map"] = lambda s, arg: s.map(arg)
+        # kwargs are forwarded to the mapper callable by pandas 3.0+ Series.map
+        self._pandas_implementations["map"] = lambda s, arg, na_action=None, **kw: s.map(
+            arg, na_action=na_action, **kw
+        )
```
chibugai, line 740: On pandas < 3.0, `Series.map` does not accept extra keyword arguments, so passing `**kw` here raises `TypeError: map() got an unexpected keyword argument ...`. In `expression_evaluator._evaluate_function_via_pandas`, that exception is silently caught and execution falls back to chDB, which cannot run a Python callable — resulting in a confusing failure rather than the mapped result. Consider a version-agnostic compat shim (consistent with `_fillna_compat` above) that binds kwargs to the callable itself, which yields identical semantics on both pandas 2.x and 3.0.
```suggestion
        def _map_compat(s, arg, na_action=None, **kw):
            # pandas 3.0+ forwards kwargs to the mapper; emulate it for older pandas
            if kw and callable(arg):
                mapper = lambda v: arg(v, **kw)
            else:
                mapper = arg
            return s.map(mapper, na_action=na_action)
        self._pandas_implementations["map"] = _map_compat
```

datastore/tests/test_pandas3_api_gaps.py
```diff
@@ -0,0 +1,232 @@
...
+        # DataStore operations (mirror of pandas)
+        ds = DataStore.from_df(pd_df)
+        ds_result = ds["a"].map(lambda v, add: v + add, add=10)
+
+        assert list(ds_result) == list(pd_result)
+        assert list(ds_result) == [11, 12, 13]
+
+    def test_map_dict_and_na_action_unchanged(self):
+        pd_df = pd.DataFrame({"g": ["A", "B", "A"]})
+        pd_result = pd_df["g"].map({"A": 4.0, "B": 3.0})
```
chibugai: Test name mentions `na_action` but the test never exercises it — it only covers the plain dict-mapping path. Since `ColumnExpr.map` was changed to merge `na_action` with `**kwargs` (`dict(na_action=na_action, **kwargs)`), a regression in `na_action` handling (e.g., a user kwarg named `na_action` colliding, or `na_action='ignore'` with NaN values) would go undetected. Add an assertion with `na_action='ignore'` on data containing NaN, or rename the test.
```suggestion
    def test_map_dict_and_na_action_unchanged(self):
        pd_df = pd.DataFrame({"g": ["A", "B", None]})
        pd_result = pd_df["g"].map({"A": 4.0, "B": 3.0}, na_action="ignore")

        ds = DataStore.from_df(pd_df)
        ds_result = ds["g"].map({"A": 4.0, "B": 3.0}, na_action="ignore")

        assert list(ds_result) == list(pd_result)
```

datastore/tests/test_pandas3_api_gaps.py
```diff
@@ -0,0 +1,232 @@
...
+
+        # DataStore operations (mirror of pandas)
+        ds = DataStore.from_df(pd_df)
+        ds_result = ds["s"].str.replace("at", "AT")
+
+        assert list(ds_result) == list(pd_result)
+
+    def test_replace_regex_form_unchanged(self):
+        pd_df = pd.DataFrame({"s": self.DATA})
+        pd_result = pd_df["s"].str.replace("[ch]at", "X", regex=True)
```
chibugai, line 172: Coverage gap: the implementation in `_build_str_replace` has a distinct code path for dict patterns combined with `regex=True` (`replaceRegexpAll` instead of `replaceAll`), but no test covers `str.replace({...}, regex=True)`. Consider adding a case like `ds["s"].str.replace({"[ch]at": "X"}, regex=True)` mirrored against pandas, so the regex branch of the new dict form isn't left untested.

datastore/function_definitions.py
```diff
@@ -248,26 +248,47 @@ def _build_replace(expr, to_replace, value=_REPLACE_SENTINEL, regex: bool = Fals
...
     """
     from .functions import Function
     from .expressions import Literal
 
+    if isinstance(pattern, dict):
+        # pandas 3.0: str.replace({'old': 'new', ...}); repl must not be given.
+        if replacement is not None:
+            raise ValueError("repl cannot be used when pat is a dictionary")
+        func_name = 'replaceRegexpAll' if regex else 'replaceAll'
+        items = list(pattern.items())
```
chibugai, line 278: Non-string dict keys/values are silently coerced via `str()`, which diverges from pandas 3.0 behavior. For example, `str.replace({'a': None})` raises `TypeError: repl must be a string or callable` in pandas, but here it would silently write the literal string 'None' into the data (similarly, numeric keys/values get stringified). This can cause silent data corruption instead of surfacing the user error. Suggest validating types up front, consistent with the `TypeError` already raised for the non-dict path.
```suggestion
        items = list(pattern.items())
        for pat, repl in items:
            if not isinstance(pat, str) or not isinstance(repl, str):
                raise TypeError("patterns and replacements in the dictionary must be strings")
```

datastore/column_expr.py
wudidapaopao: Adopted in 5e9017090f5 — call-site capability check (VAR_KEYWORD in `Series.map`'s signature) raising `NotImplementedError`, consistent with the iceberg additions. A new pandas-2 test asserts the error fires eagerly at the call site.

datastore/pandas_compat.py
wudidapaopao: Adopted in 5e9017090f5 — dropped the function-level import; the method now uses the module-level `pd` like the other `to_*` methods in this mixin.

datastore/function_executor.py
wudidapaopao: Kept the pass-through, addressed the failure mode differently: with the eager call-site check added per your other comment, kwargs can no longer reach this lambda on pandas < 3, so the swallowed-TypeError → chDB-fallback confusion can't trigger. I preferred that over the binding shim for two reasons: (1) parity over enhancement — kwargs silently working on pandas 2 would diverge from what the same code does on plain pandas; (2) the shim would mis-bind pandas 3's `engine` keyword (an explicit `Series.map` parameter, not forwarded to the callable) into the mapper. Keeping `**kw` pass-through preserves native pandas-3 semantics including `engine=`.

datastore/tests/test_pandas3_api_gaps.py
wudidapaopao: Adopted in 5e9017090f5 — split into `test_map_dict_mapping_unchanged` plus a new `test_map_na_action_ignore_with_missing_values` using your suggested data, mirrored against pandas with the NaN position asserted explicitly.

datastore/tests/test_pandas3_api_gaps.py
wudidapaopao: Adopted in 5e9017090f5 — added a pandas-3 mirror test for the dict+regex branch with {'[ch]at': 'X', 'X+': 'Y'} (which also exercises sequential regex application) and a version-agnostic value test against chained `re.sub`.

datastore/function_definitions.py
wudidapaopao: Adopted in 5e9017090f5 — dict entries are validated up front and raise `TypeError` like the non-dict path; the `str()` coercion is gone. Tests cover {'cat': None} and {1: 'one'}.

wudidapaopao:

ShawnChen-Sirius (approved):
LGTM, approve to merge.

ShawnChen-Sirius:

PR #618, opened 2026-07-27
Replay WAL in the object's database context

ShawnChen-Sirius:
## What

Restore the object's database as the current database before replaying the write-ahead log in `chdb.durable`.

## The bug

WAL replay ran with `default` as the current database. A durable object sets its own database (the manifest's `db`) as current, and callers write with unqualified table names against it. On reopen, those logged statements replayed into `default` instead — so the tables were invisible to the caller and dropped by the next `BACKUP DATABASE {db}` checkpoint. Net effect: **committed writes silently lost after a crash/reopen**, whenever the caller used unqualified names (the natural style once you've set a database).

## The fix

`USE {db}` before replay (one line in `_restore`), plus a note in the WAL usage rules.

## Test

Adds a regression test. Every existing durable test qualifies its table names (`mem.t`), so none exercised the replay database context; the new test uses an unqualified name and reopens. Without the fix it fails with `UNKNOWN_TABLE`; with it, the full suite passes (verified locally and against MinIO).

## Context

Found while building a memory backend on `chdb.durable` (auxten/clickmem#7), whose tables are unqualified — this fix is a correctness prerequisite for that use.

> [!NOTE]
> ### Replay WAL in the object's database context and quote database identifiers
> - Adds `_quote_ident` helper in [object.py](https://github.com/chdb-io/chdb/pull/618/files#diff-65082cd9f06eff60433e3c7c536498cd1971cf7610af29e5ff6e7584c1aba610) to backtick-quote ClickHouse identifiers, allowing database names containing hyphens or other special characters in DDL statements.
> - Issues a `USE <db>` statement before WAL replay in `DurableObject._restore` so unqualified table references in WAL entries replay against the object's own database instead of `default`.
> - Applies `_quote_ident` consistently in `open`, `_restore`, and `checkpoint` for `CREATE DATABASE` and `BACKUP DATABASE` statements.
> - Behavioral Change: WAL replay now executes in the object's database context; previously unqualified statements would silently target `default`.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 9cd8e66.</sup>
> 
>

### DIFF
--- a/chdb/durable/object.py
+++ b/chdb/durable/object.py
@@ -44,6 +44,14 @@ def validate_oid(oid: str) -> str:
     return oid
 
 
+def _quote_ident(name: str) -> str:
+    """Backtick-quote a ClickHouse identifier so a database name with `-` (or
+    any character needing quoting) is valid in DDL/USE, and can't break out of
+    the statement. Not imported from clickhouse_connect to keep this package
+    free of that dependency."""
+    return "`" + name.replace("\\", "\\\\").replace("`", "\\`") + "`"
+
+
 class DurableObject:
     def __init__(self, oid: str, backend: Backend, *, owner: Optional[str] = None,
                  db: str = "mem", read_only: bool = False, lease_ttl: float = 60.0):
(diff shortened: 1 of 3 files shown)


ShawnChen-Sirius:
@chibugai please review it.

ShawnChen-Sirius:
@wudidapaopao please review it.

wudidapaopao (approved):
LGTM

PR #627, opened 2026-08-19
Push DataStore SQL segments to the server that owns the data

ShawnChen-Sirius:
Sends a DataStore chain's compiled SQL to the ClickHouse server that owns the table, instead of pulling its rows into the local engine, and reports where every segment of a plan ran.

- The data source is rendered for the engine that will run the segment: SQL bound for the server reads `database.table` directly, never `remote()`.
- A bound segment executor decides which sources it can serve; a remote failure is raised, never retried locally.
- Foreign results are decoded by the engine from ClickHouse's Native bytes, so a pushed-down result carries the dtypes local execution produces.
- Every execution publishes where each segment ran and why, including the segments that stay local, with stable reason codes.

Nothing changes without an executor bound: the compile target stays local, the observer is unset, and existing paths are untouched. `datastore` suite: 11242 passed, 104 skipped, 81 xfailed, 4 xpassed.

Consumed by chdb-io/generic-compute#1, which verifies the shared contract against a live ClickHouse server.

Supersedes #626, closed when its branch was renamed to drop the `codex/` prefix.

> [!NOTE]
> ### Push DataStore SQL segments to the remote ClickHouse server that owns the data
> - Adds a `SqlSegmentExecutor` abstract interface in [pushdown.py](https://github.com/chdb-io/chdb/pull/627/files#diff-a3c381b9c7e8df4563e2c10fb84be943f54e4d036a8911f2b0ab42615d7cf156) that callers implement to run SQL on a remote ClickHouse server; bind one to a `DataStore` via `set_sql_segment_executor()`.
> - When a compatible executor is bound and the data source is a `remote()` table function, the first SQL segment is compiled as a direct table reference (no `remote()` loopback) and executed on the owning server, with the result returned as a DataFrame for the remainder of the pipeline.
> - Adds `PushdownTrace` so callers can inspect the target, SQL, source, row count, and metrics after remote execution via `last_pushdown_trace`.
> - Introduces `PushdownReasonCode` and `PushdownDecision` in [query_planner.py](https://github.com/chdb-io/chdb/pull/627/files#diff-14581faecfe6f0baf2758354cd0328729c05136f97da1304ad89831f8d1d0870) with an `explain()` method on `ExecutionPlan`, giving structured per-operation pushdown decisions with stable reason codes.
> - Adds a global plan observer hook (`set_plan_observer`) that receives `SegmentPlacement` reports after each execution without affecting query results if the observer raises.
> - Risk: remote execution failures are not retried locally; a failed remote segment surfaces as an error rather than falling back to chDB.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized a07b40a.</sup>
> 
>


ShawnChen-Sirius:
Closing: this work now lives on `test/generic-compute`, alongside the UDF deploy branch. generic-compute pins one chdb, so its build cannot choose between the two features and they have to travel together. Every commit that was here is in that branch — verified with `git log --cherry-pick`, which reports nothing exclusive to this one. Deleting the branch loses no work.

PR #636, opened 2026-09-08
Add a chdb console script so pipx and uvx can run the CLI

### LINKED ISSUE #125: Include an entrypoint for compatibility with pipx
I'd like to be able to do this:

```bash
pipx install chdb
chdb "SELECT 1,'abc'" Pretty
```
As an alternative to this:
```bash
pip install chdb
python3 -m chdb "SELECT 1,'abc'" Pretty
```

The reason I want this is that `pipx` automates the process of creating a separate virtual environment for each tool, so it lets me try out tools like `chdb` without manually creating a virtual environment for it.

This currently doesn't work, because the `chdb` package does not declare a CLI tool entrypoint. I get this error:

```bash
pipx install chdb
```
```
No apps associated with package chdb or its dependencies. If you are
attempting to install a library, pipx should not be used. Consider using pip
or a similar tool instead.
```
I think one way to implement this would be to add this to `setup.py`:

```python
    entry_points="""
        [console_scripts]
        chdb=chdb.__main__:main
    """,
```

wudidapaopao:
Closes #125.

chdb ships a CLI (`chdb/__main__.py`, from chdb-core) but the wheel declares no
entry point, so it is reachable only through `python -m chdb`. That rules out
the zero-setup installers people reach for first:

```
$ pipx install chdb
No apps associated with package chdb ...
$ uvx chdb "SELECT 1"
error: Failed to spawn: `chdb`
```

One `[project.scripts]` entry fixes both, and `uvx chdb "SELECT 1"` becomes a
one-liner with no virtualenv to manage.

Verified with a wheel built from this branch, installed next to
chdb-core 26.7.0 on Python 3.12:

```
$ chdb "SELECT 1,'abc'" Pretty
   ┏━━━┳━━━━━━━┓
   ┃ 1 ┃ 'abc' ┃
   ┡━━━╇━━━━━━━┩
1. │ 1 │ abc   │
   └───┴───────┘
$ chdb --help | head -1
usage: chdb [-h] "SELECT 1" [format]
```

Needs chdb-io/chdb-core#219 to be released alongside it: `main()` currently
raises `NameError: name '__path__' is not defined` (the same crash that has made
`python -m chdb` unusable since 4.1.8), so the console script would land broken
without it.

> [!NOTE]
> ### Add `chdb` console script entry point in `pyproject.toml`
> Registers a `chdb` project script that dispatches to `chdb.__main__:main`. This lets pipx and uvx run the CLI as a standalone command after install.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 588abab.</sup>
> 
>

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -52,8 +52,6 @@ durable = ["boto3>=1.36"]  # IfMatch/IfNoneMatch conditional writes on PutObject
 durable-gcs = ["google-cloud-storage"]
 durable-azure = ["azure-storage-blob"]
 
-# Expose the `python -m chdb` CLI as a `chdb` command, so `pipx install chdb` and
-# `uvx chdb "SELECT 1"` work. main() lives in chdb-core's chdb/__main__.py.
 [project.scripts]
 chdb = "chdb.__main__:main"
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -52,6 +52,11 @@ durable = ["boto3>=1.36"]  # IfMatch/IfNoneMatch conditional writes on PutObject
 durable-gcs = ["google-cloud-storage"]
 durable-azure = ["azure-storage-blob"]
 
+# Expose the `python -m chdb` CLI as a `chdb` command, so `pipx install chdb` and
+# `uvx chdb "SELECT 1"` work. main() lives in chdb-core's chdb/__main__.py.
+[project.scripts]
+chdb = "chdb.__main__:main"
+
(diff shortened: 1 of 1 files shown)


wudidapaopao:
@ShawnChen-Sirius plz review this PR.

ShawnChen-Sirius (approved):
LGTM, approve to merge.

ShawnChen-Sirius:

PR #635, opened 2026-09-08
Bring chdb.durable onto the frozen Durable V1 contract

ShawnChen-Sirius:
`chdb.durable` predates the frozen V1 protocol, and chdb-core v26.7.2-rc.2 now exports the ABI that protocol assumes. This brings the Python binding onto both.

**Engine, not string handling.** Backup and restore call `chdb_backup_database_n` / `chdb_restore_database_n`; nothing here assembles `BACKUP` or `RESTORE` text. Every public statement goes through `chdb_classify_query_n` — `query()` takes one `READ_ONLY` statement, `execute()` takes one `MUTATING` statement with no credential that writes only into this object's database and does not change its container. Refusals never quote the SQL.

**The frozen object format.** `head.json` gains `protocol` and `engine` sections; base and WAL references carry length + SHA-256, verified before use; a released lease has one fixed shape; unknown fields round-trip. Opening compares `backup_format` and `min_reader`, not version equality.

**Also.** The 14 error categories as distinguishable types; a per-object operation lock the lease heartbeat queues behind; self-fencing on unconfirmable renewal; §5.8 reconcile that reports `commit_ambiguous` rather than unproven success; checkpoints uploaded/downloaded by file.

Engine identity comes from `SELECT chdb()`, which returns the same `CHDB_VERSION` constant the C `chdb_version()` returns — that function is exported but not bound into Python, and `engine_version()` will prefer the binding if a later core adds one.

**Tests** are rebuilt as the contract's §7.2 fixtures, §7.3 scenarios, and §7.4 fault matrix (49 checks), driven partly by a backend that lies about what the provider said. `ALL PASS` against a locally built v26.7.2-rc.2 core.

**Two things reviewers should weigh in on.**

- `pyproject.toml` still says `chdb-core>=26.7.0`, and only `v26.7.2-rc.2` carries the ABI — a pre-release pip will not pick by default. So on a default install `chdb.durable` refuses at `open()` and this test module skips. Raising the floor needs a stable core release first; not done here.
- Objects written by the pre-V1 prototype are refused as `corrupt` (their checkpoints carry no length or digest, so a V1 reader cannot verify them). The error says so and points at export-and-reload. Community workaround is being tracked separately.

Out of scope, tracked as follow-ups: cross-binding fixture exchange (§7.5) and GC (§8.1). `destroy()` stays as a development convenience, marked as outside V1.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

> [!NOTE]
[...]

PR #637, opened 2026-09-09
Run the durable conformance suite against an engine that has the ABI

ShawnChen-Sirius:
The durable conformance step added with the V1 contract (#635) was green whether or not it ran.

No chdb-core on PyPI exports `backup_database` / `restore_database` / `classify_query` — 26.7.0 is the newest published — so resolving the dependency normally left `tests/test_durable.py` skipping itself at module level, and the step accepted pytest's exit 5 to let that pass. The 49 checks have therefore never run anywhere but the machine they were written on.

The wheels carrying the ABI are already attached to the [chdb-core v26.7.2-rc.2 release](https://github.com/chdb-io/chdb-core/releases/tag/v26.7.2-rc.2), so this pins one per platform and installs it directly. No build, no PyPI release, one download per job.

**Two things about it are less obvious than they look.**

The engine install is followed by a second install of the wrapper wheel, which is not redundant: `chdb` and `chdb-core` both own `chdb/__init__.py`, so whichever lands last wins. Measured, the two orders really do differ — engine-last leaves chdb-core's 10594-byte copy in place instead of the wrapper's 8366-byte one. Both pass today only because chdb-core ships a compatible superset, and that is not a property to build a test job on. `--no-deps` on both installs keeps pip from resolving chdb-core back down to what PyPI offers.

The exit-5 tolerance is **removed** rather than kept as a safety net. It existed to excuse the expected skip; with the engine pinned there is nothing left for it to excuse, and collecting nothing now means the suite stopped running. A one-line assertion ahead of pytest names the engine actually loaded, so a wrong pin reads as one line instead of 49 collection errors.

**Verification** (macOS arm64, against the release wheel this pins rather than a local build):

```
ABI present: True
engine_version(): 26.7.2-rc.2
49 passed in 3.94s
```

All four pinned URLs were checked against the release's asset list, and each workflow gets its own platform's wheel.

**The pin is deliberately a pre-release**, so nothing here reads as a stable dependency. It comes out when a chdb-core carrying the ABI is what pip resolves by default and chdb's floor rises to match.

Not in scope: `_ABI_HINT` in `chdb/durable/engine.py` still says `pip install -U chdb-core`, which lands on 26.7.0 and helps nobody. Fixing it means naming an install channel that does not exist yet, so it belongs with that channel rather than here.

> [!NOTE]
> ### Run durable conformance suite against pinned chdb-core 26.7.2 wheel in CI workflows
> - Installs the architecture-specific `chdb-core` 26.7.2 release-candidate wheel before reinstalling the built `chdb` wrapper in the Linux arm64/x86_64 and macOS arm64/x86_64 wheel workflows
> - Adds a preflight check that fails the step when the Durable V1 ABI is missing and prints the engine version
> - Runs the durable pytest suite directly so its real exit status propagates, including exit 5 for no collected tests, instead of tolerating skips or empty runs
> - Removes the wheel-workflow glob from the push and pull_request `paths-ignore` filters so edits to these workflow files can trigger the workflows
> - Risk: the durable step now fails on missing Durable V1 ABI or empty test collection where it previously passed; reviewers should check the preflight check and pytest invocation in each modified workflow file
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized dbf2517.</sup>
> 
>

### DIFF
--- a/.github/workflows/build_linux_arm64_wheels-gh.yml
+++ b/.github/workflows/build_linux_arm64_wheels-gh.yml
@@ -194,29 +194,48 @@ jobs:
           python -m pip install dist/*.whl --force-reinstall
           python -m pip install pytest
 
-          # Durable V1 needs chdb-core's backup / restore / query-analysis ABI.
-          # Until a chdb-core carrying it is what pip resolves by default, the
-          # module skips itself — and that skip is as much the thing under test
-          # as the checks are, so this step is green either way and turns into
-          # real coverage the day the dependency catches up.
-          set +e
-          python -m pytest tests/test_durable.py -v --tb=short
-          code=$?
-          set -e
-          if [ "$code" -eq 5 ]; then
-            # Exit 5 is pytest's "collected nothing", which is what a
-            # module-level skip looks like from outside. Accepting it blindly
-            # would also swallow a suite that stopped being collected for some
-            # unrelated reason, so make it prove why it skipped.
-            python - <<'PY'
+          # Move the engine to the pre-release that carries the Durable V1 ABI,
+          # then put the wrapper back on top of it.
+          #
+          # The second install is not redundant. chdb and chdb-core both own
+          # chdb/__init__.py, so whichever lands last wins, and the wrapper's
+          # is the one that has to: chdb-core ships a superset that happens to
+          # work today, which is not a property to build a test job on.
+          #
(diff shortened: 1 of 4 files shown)


ShawnChen-Sirius:

wudidapaopao (approved):
LGTM

PR #640, opened 2026-09-10
Give the bindings one local URL and one default database

ShawnChen-Sirius:
Two things the same durable object needed different code for depending on which language opened it, and neither difference was anything the contract asked for.

**One URL.** The local backend answered only to `local:`, chdb-node only to `file:`, chdb-go to either — so no single namespace URL could be handed to all three. It now takes both here (chdb-node takes both as of chdb-io/chdb-node#94). A `file:` URL is percent-encoded and a `local:` path is not; that is the only difference between them.

**One default database.** A cold object was created in `mem` here and `default` in Node and Go. The head is authoritative once an object exists, so this never broke interop — but the same application ported between bindings addressed a differently-named table. `default` wins because two of three bindings already had it and it is what an unqualified statement means in chDB anyway. The contract now states it, so there is one answer rather than three implementations.

**The refusal says which engine is loaded.** It named the ABI it needed and nothing else, so its two causes read identically: an engine too old to have the ABI, and a new engine that the wrapper is not loading because something else owns `chdb/`. And the command it printed only became true with chdb-core 26.7.3 — the first release carrying the ABI that is published to PyPI. It names `chdb-core` rather than `chdb` because pip leaves a satisfied dependency alone, so upgrading `chdb` reports success and moves no engine.

```
chdb.durable needs chdb-core 26.7.2-rc.2 or newer, which added backup_database / restore_database / classify_query.
The engine loaded here reports 26.7.0.

    pip install -U chdb-core
```

## Verified

**Cross-binding recovery, contract §7.5 — the item no repository had implemented.** Every writer produces a 100-row checkpointed base plus a 50-row WAL tail that is flushed and not checkpointed, so a reader has to restore the base *and* replay the WAL. 3×3, on the local backend and again on real AWS S3, every cell 150 rows / `sum(id)` 11175 / `base` 100 + `wal` 50, each reader taking the database name from `head.json`:

| writer \ reader | Python | Node | Go |
|---|---|---|---|
| Python | ✓ | ✓ | ✓ |
| Node | ✓ | ✓ | ✓ |
| Go | ✓ | ✓ | ✓ |

After this change all three take the same URL string and the same database name.

Also: `tests/test_durable.py` 50 passed against chdb-core 26.7.2-rc.2, and the whole suite again against real AWS S3 (`DAO_TEST_URL=s3://…`, 49 passed in 7 minutes). The refusal message was checked from where a user actually is — `chdb` on PyPI's newest chdb-core, 26.7.0 — and it names 26.7.0.

## Two tests changed rather than being added to

`test_control_statements_refused` and `test_reopen_honors_persisted_db` both relied on the default being `mem`. With the default moved, the first one's `USE default` would have been a no-op instead of a refusal, and the second would have had nothing to disagree with. Both now name `mem` explicitly, which is what they were really testing.

## Follow-ups, once 26.7.3 is on PyPI

- add `chdb-core>=26.7.3` to the `durable` extra, so `pip install "chdb[durable]"` resolves an ABI-carrying engine deterministically
- drop the `CHDB_CORE_ABI_WHEEL` pin from the durable conformance step in the four wheel workflows; it exists only because no chdb-core on PyPI exported the ABI

> [!NOTE]
> ### Standardize durable bindings on `default` database and add `file:` URL support to `make_backend`
> - Adds the shared `DEFAULT_DATABASE` constant (value `default`) in [protocol.py](https://github.com/chdb-io/chdb/pull/640/files#diff-2d45447a0277247a514842e3c9231e6ac99bea23a40ceb7293ac058af06cf364) and switches `DurableObject`, `Namespace`, and contract docs from the old `mem` default to it. Persisted head data remains authoritative on reopen.
> - Updates `make_backend` in [backends/__init__.py](https://github.com/chdb-io/chdb/pull/640/files#diff-edf2a4593a6ca5d237c043410bc0506bca8ddfaabb6e878934df53663d42288a) to route `file:` URLs into the local filesystem backend, percent-decoding the path before resolution; `local:` and bare paths stay literal.
> - Improves ABI mismatch errors in [engine.py](https://github.com/chdb-io/chdb/pull/640/files#diff-b6f345c95c93f34ee8ac5e9ebd4fad6c0f9fde56da4d6c0414ccd84a70e60d48) to report the loaded engine version and, for `ManagedConnection`, the missing required attributes.
> - Adds installation docs and bumps the documented minimum `chdb-core` to 26.7.2-rc.2.
> - Behavioral Change: cold-created durable objects without an explicit `db` now land in `default` instead of `mem`; existing tests that relied on the implicit `mem` default are updated to pass it explicitly.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 0dad0fc.</sup>
> 
>

### DIFF
--- a/chdb/durable/__init__.py
+++ b/chdb/durable/__init__.py
@@ -25,7 +25,7 @@ buffers it; `flush()` is what makes it recoverable elsewhere. A service that
 promises a completed request survives a crash has to await `flush()` before it
 answers.
 
-Requires a chdb-core with the Durable V1 ABI (26.7.2-rc.2 or newer): backup,
+Requires chdb-core 26.7.2-rc.2 or newer (`pip install -U chdb-core`): backup,
 restore and statement classification are the engine's job, and this package
 builds no `BACKUP`/`RESTORE` SQL and runs no regex over a caller's statement.
 
--- a/chdb/durable/backends/__init__.py
+++ b/chdb/durable/backends/__init__.py
@@ -30,7 +30,7 @@ state machine reconciles it (§5.8) into success, `lease_fenced`, or
 `Namespace.destroy`, which is a development convenience outside the protocol.
 
 `make_backend(url, sub)` maps a URL to a backend scoped to `<base>/<sub>`:
-    local:/path/to/root
+    local:/path/to/root         (`file:///path/to/root` and a bare path too)
     s3://bucket/prefix          (endpoint via CHDB_DURABLE_S3_ENDPOINT for MinIO/R2)
     gcs://bucket/prefix
     azure://container/prefix
@@ -39,7 +39,7 @@ from __future__ import annotations
 
 import os
 from typing import Optional, Protocol, runtime_checkable
-from urllib.parse import urlparse
+from urllib.parse import unquote, urlparse
 
 
 @runtime_checkable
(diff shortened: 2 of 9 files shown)


ShawnChen-Sirius:

wudidapaopao (approved):
LGTM

